PR #1, opened 2016-12-23
Update Dockerfile

KalleDK:
Adding root / to /var/tmp


ziggimon:
Fixed with: https://github.com/plexinc/pms-docker/commit/8eda80d6c3415f1fdbd969a6793b4b15be9027cc

Thanks Kalle 😄

PR #2, opened 2016-12-23
Add `.env` support for docker-compose

xFlatlinex:
Docker-compose supports setting variables using a `.env` file, so
instead of having to modify the original YAML file, this can be used
instead.

A sample `.env.example` is provided so the user can use it as a guide.
The `.env` file is ignored, as it's content is only relevant to the
user.


gbooker:
I'm not sure it's practical to expect the user to not have to modify the original YAML file.  The structure for the mounts for media is almost certainly going to require modification for individual users no matter what the default template has in it since users have such wildly varying media layouts.  This is evident by the fact that your commit includes envs that are examples but almost certainly not applicable to the end user (`VOL_MEDIA_A`, etc…).

Given that, what significant advantage does the `.env` file give?

KalleDK:
I see no advantages for the `.env` file as you still would have to change the mountpoints. I would though suggest renaming the `docker-compose.yml` to `docker-compose.yml.example`, to suggest that this is a template and would not work out of the box. (And add `docker-compose.yml` to .gitignore, so you can have your the docker-compose.yml with the repo)

gbooker:
`docker-compose.yml` moved to `docker-compose.yml.template` and appropriate ignore files added.  This should satisfy the spirit of this PR more than the original.

xFlatlinex:
Thanks @gbooker, didn't realize at the time that those were arbitrary/user-defined. So @KalleDK's solution is the way to go. I'm closing this PR.

PR #8, opened 2016-12-24
fix permission detection

aptalca:
If PLEX_UID is set to an existing user's uid, like 0 (root), then `stat -c %U` returns "root" instead of "plex" and forces an unnecessary `chown -R`. Instead using `stat -c %u` and checking it against the PLEX_UID is more accurate and prevents unnecessary chowns.

### DIFF
--- a/root/etc/cont-init.d/40-plex-first-run
+++ b/root/etc/cont-init.d/40-plex-first-run
@@ -50,7 +50,7 @@ fi
 # Update ownership of dirs we need to write
 if [ "${CHANGE_CONFIG_DIR_OWNERSHIP,,}" = "true" ]; then
   if [ -f "${prefFile}" ]; then
-    if [ ! "$(stat -c %U "${prefFile}")" = "plex" ]; then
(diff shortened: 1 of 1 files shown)


aptalca:
capture33

Here's a look from inside the container when PLEX_UID is set to "0"

gbooker:
I agree that doing a numeric comparison is better, but perhaps instead of `$(PLEX_UID}` (which may not be set), a `$(id -u plex)` would be better?  At this stage of execution, if `PLEX_UID` was set, the `plex` user has already had its id changed.

KalleDK:
You could also add `PLEX_UID=${PLEX_UID:-$(id -u plex)}` in the top, and thereby keep consistency in the script

aptalca:
My bad, didn't realize that PLEX_UID is optional. 

I agree that `$(id -u plex)` is the less intrusive option. I'll update the PR

gbooker:
Merged in 5413de3442029e782885af2030a419d5a9ba53d8

PR #12, opened 2017-01-04
Add healthcheck using web UI

double16:
Docker 1.12 added the ability to specify a health check via the HEALTHCHECK instruction. Docker 1.12 is required to build with the HEALTHCHECK instruction, but older versions of docker can run the image, the health check information is ignored.

This PR adds a health check based on http://localhost:32400/web/index.html returning a successful status code.

Is there a better URL for status? Multiple URLs for different services?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,3 +50,6 @@ COPY root/ /
 RUN \
 # Save version and install
     /installBinary.sh
+
+HEALTHCHECK --interval=200s --timeout=100s CMD /healthcheck.sh || exit 1
+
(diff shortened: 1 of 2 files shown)


gbooker:
A better URL would be `/identity` as it is served up without a token even when security is enabled and it is not serving up a file.  Also there are a few rare circumstances where the embedded web interface files are missing and the web fetch will fail.

gbooker:
Merged: @3ae89855b41214004fd2f25a212d771dd9757d60

PR #14, opened 2017-02-12
Patch 1

bxnyintuc:


ziggimon:
I don't believe these changes are taking the entire specter of users into account.

1: Users do live in other time zones than Pheonix. That's why we use the wording `timezone`
2: Most users will be on Linux like systems and will not have C:\ paths. Also why we wrote the comments in wording that most if not all can understand.

So honestly I don't see how this PR brings any constructive changes to help the entire Plex on Docker user base and I don't think we will merge this. A better PR would be to add wording to the explaination of the `-v` flag to assist windows users.

ziggimon:
Closing based on previous comment.

PR #17, opened 2017-04-22
Ensures 'tzdata' package is installed.

yorkedork:
This package is necessary in order to provide zoneinfo and support
for using the TZ environment variable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN \
 # Update and get dependencies
     apt-get update && \
     apt-get install -y \
+      tzdata \
       curl \
       xmlstarlet \
       uuid-runtime \


gbooker:
Has something changed recently?  The images we ship have `tzdata` in them, but when running through the docker commands to create the image starting with a base `ubuntu:16.04`, it does not have `tzdata`.

yorkedork:
If it helps, I can confirm that I'm building the image locally and that I tzdata is not installed. Of course, I can't comment on your build process or artifacts that are being pushed to docker hub.

gbooker:
Our build process is quite simple: `docker build --build-arg TAG=$VERSION -t plexinc/pms-docker:$VERSION .`  What I mean is has something changed in Ubuntu such that the same commands are no longer installing `tzdata`?

yorkedork:
It would appear that the cloudimage tarballs (https://partner-images.canonical.com/core/xenial/) that are used to generate the official ubuntu docker images had the tzdata package installed _before_ 2017-04-10, but not after.

The docker image repository was updated on 2017-04-12 (https://github.com/tianon/docker-brew-ubuntu-core/commits/94a6b954435615aba4beffaf62f14699083eafca/xenial/Dockerfile) to use the newer tarballs that don't appear to have this package installed.

Please note that I'm making this determination by just opening the archives and looking for the presence of `/usr/share/zoneinfo`.

gbooker:
I locally confirmed that the latest `ubuntu:16.04` image does not contain the `tzdata` package.  Our CI is caching a previous `ubuntu:16.04` which still contains `tzdata` so the images we pushed to docker-hub are fine.

PR #18, opened 2017-05-08
Advice for known issues with headless server setup

raubreywhite:
If you try to access http://ip_of_server:32400/web, the server will not be found. It can only be "found" (and subsequently setup) by accessing it from http://localhost:32400/web. A SSH tunnel removes the need to plug your headless server into a monitor+keyboard+mouse.


ziggimon:
I see no harm in letting users know about this challenge. Over the past its been covered immensely, but new users would perhaps not know. So adding for better clarification of headless use. Thanks!

ziggimon:
Merged with: 5f68c80c9ea0fc7dafdb4bcc72b9fbc63f77d739

PR #19, opened 2017-06-10
s6: v1.19.1.1

MindTooth:
Shipped version is far behind.


gbooker:
What is the appreciable difference?  Aside from feature additions which we don't really need, I don't see any.

MindTooth:
I had hoped that this might solve my problem with stalling containers.  Plex seems to stall my system on shutdown.

But I'll guess you are right.  Feel free to close.

gbooker:
OK, since you can reproduce an issue and think this may solve it…

- Clone the repo
- Make your change
- Build and image
- Try the new image and see if it solves anything.

MindTooth:
screen shot 2017-06-15 at 22 27 05

Well, that did nothing :-P

PR #20, opened 2017-07-16
Update Dockerfile

erikrz:
Add data volume, since the directory is already created


gbooker:
The '/data' is left out of the list of volumes intentionally.  Per the documentation, the `/data` volume in the examples are provided as an example only and it is up to the user to do their preferred data structure layout.

PR #23, opened 2017-10-04
added data volume to dockerfile

CleverCompiler:


gbooker:
From https://github.com/plexinc/pms-docker#parameters

> This [volume] is provided [an example] for providing media into the container. The exact structure of how the media is organized and presented inside the container is a matter of user preference. You can use as many or as few of these parameters as required to provide your media to the container.

Therefore, the `/data` is an optional volume and users could present their media to the container using any number of volumes of which none may be at `/data`.  The `/transcode` and `/config` are both explicitly used in the container where as `/data` would only be used if the user added a library that used that path.

PR #29, opened 2017-12-04
Change connection timeout to overall timeout. 

midzelis:
In case connection succeeds, but no data returned.

### DIFF
--- a/root/healthcheck.sh
+++ b/root/healthcheck.sh
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 
 TARGET=localhost
-CURL_OPTS="--connect-timeout 15 --silent --show-error --fail"
+CURL_OPTS="--max-time 15 --silent --show-error --fail"
 
 curl ${CURL_OPTS} "http://${TARGET}:32400/identity" >/dev/null
 


gbooker:
The `HEALTHCHECK` in the Dockerfile already has an overall timeout.  What situation is this supposed to cover that was not already done?

midzelis:
It was leaving a lot of old curl processes around. This was happening due to trying to run /config on a NFS share, which isn't supported due to sqlite file locking issues. It would hang the webserver/plex server. The healthcheck could connect to the webserver, but there was no response, leaving the curl processes hanging around too. This will make the healthcheck more responsive.

gbooker:
In the case you describe, did the healthcheck fail or did it pass when it should not have done so?

Also, 15 seconds is a bit short for an overall timeout.  There are some rare processes (such as database optimize) that can lock up the database for that long or longer on low end hardware (such as NAS devices).

gbooker:
Closing due to lack of response.

midzelis:
healthcheck didn't fail, didn't pass either. Just created new processes every X minutes. (didn't clean up processes either, strangely - maybe a docker healthcheck bug?)

I don't have a problem anymore, since I moved /config off of the bind mount (NFS) docker share, and using docker volume instead.

PR #30, opened 2018-02-24
Resetting ownership of mounted paths at restart

ianmaddox:
There are cases when the ownership of the transcode dir is reset to root when starting the container. This fix runs the logic to fix the ownership of /config and /transcode at every restart.

### DIFF
--- a/root/etc/cont-init.d/50-plex-update
+++ b/root/etc/cont-init.d/50-plex-update
@@ -7,6 +7,9 @@ fi
 
 . /plex-common.sh
 
+# Ensure proper transcode directory ownership
+chown -R plex:plex /transcode
+
 function getPref {
   local key="$1"
 
@@ -54,15 +57,3 @@ fi
 # Do update process
 echo "Attempting to upgrade to: ${remoteVersion}"
 installFromUrl "${remoteFile}"
-
-# Update ownership of dirs we need to write
-if [ "${CHANGE_CONFIG_DIR_OWNERSHIP,,}" = "true" ]; then
-  if [ -f "${prefFile}" ]; then
-    if [ ! "$(stat -c %u "${prefFile}")" = "$(id -u plex)" ]; then
-      chown -R plex:plex /config
-    fi
-  else
-    chown -R plex:plex /config
-  fi
-  chown -R plex:plex /transcode
-fi
--- a/root/etc/cont-init.d/50-plex-update
+++ b/root/etc/cont-init.d/50-plex-update
@@ -54,3 +54,15 @@ fi
 # Do update process
 echo "Attempting to upgrade to: ${remoteVersion}"
 installFromUrl "${remoteFile}"
+
+# Update ownership of dirs we need to write
+if [ "${CHANGE_CONFIG_DIR_OWNERSHIP,,}" = "true" ]; then
+  if [ -f "${prefFile}" ]; then
+    if [ ! "$(stat -c %u "${prefFile}")" = "$(id -u plex)" ]; then
+      chown -R plex:plex /config
+    fi
+  else
+    chown -R plex:plex /config
+  fi
+  chown -R plex:plex /transcode
+fi


gbooker:
> There are cases when the ownership of the transcode dir is reset to root when starting the container.

Such as?

ianmaddox:
Every time the container restarts.

`dr-xr-xr-x 636 root root    0 Mar  4 21:14 proc
drwxr-xr-x   2 root root   40 Mar  4 21:14 transcode
drwxr-xr-x   5 root root  340 Mar  4 21:14 dev
drwxrwxrwt   3 root root 4.0K Mar  4 21:14 tmp
root@plex:/#`

Natteverf:
It sounds to me like you are not using the PLEX_UID and PLEX_GID parameters or that you have them set incorrectly:
These parameters are usually not required but some special setups may benefit from their use. As in the previous section, each is treated as first-run parameters only:

    PLEX_UID The user id of the plex user created inside the container.
    PLEX_GID The group id of the plex group created inside the container

Users/Groups

Permissions of mounted media outside the container do apply to the Plex Media Server running within the container. As stated above, the Plex Media Server runs as a specially created plex user within the container. This user may not exist outside the container and so the PLEX_UID and PLEX_GID parameters are used to set the user id and group id of this user within the container. If you wish for the Plex Media Server to run under the same permissions as your own user, execute the following to find out these ids:

$ id `whoami`

You'll see a line like the following:

uid=1001(myuser) gid=1001(myuser) groups=1001(myuser)

In the above case, if you set the PLEX_UID and PLEX_GID to 1001, then the permissions will match that of your own user.

ianmaddox:
You are correct. Once I added those, it works fine.
This feature isn't well documented on the docker side. I'll add a note to the readme unless someone beats me to it. It's still worth adding a patch for this problem, or at least a check in the health check script, as Plex isn't very useful without the transcode dir.

gbooker:
It's already in the readme.  Closing due to being a non-issue.

dstar4138:
I'd actually like if we could revisit this. 

TL;DR: Mounting `/transcode` as a tmpfs will mount it as root: (https://github.com/docker/compose/issues/3270). This means no matter what I set UID/GID, it will not work.

In my current scenario, I'm running docker on a NAS with an intentionally small `/tmp` and `swap`. It isn't practical to map `/transcode` to the global `/tmp`. But I have plenty of ram, so a unique tmpfs for the container is optimal right? Thus, I'm left with this snippet from my docker-compose.yml.

```yaml
services:
    pms:
        # To perform updates, simply restart the container.
        image: plexinc/pms-docker:plexpass
        restart: unless-stopped

        # Only the HTTP PMS port is required.
        # All others are optional (but 1900 is needed for most apps).
        ports:
            - 32400:32400/tcp  # http PMS access
            - 1900:1900/udp    # DLNA Server
            - 32469:32469/tcp  # DLNA Server
            - 3005:3005/tcp    # Home Theater control via Plex Companion. 
            # 8324:8324/tcp    # Plex for Roku via Plex Companion.
            - 32410:32410/udp  # GDM Network Discovery #1
            - 32412:32412/udp  # GDM Network Discovery #2
            - 32413:32413/udp  # GDM Network Discovery #3
            - 32414:32414/udp  # GDM Network Discovery #4
        environment:
            # Go here for token: https://www.plex.tv/claim
            - PLEX_CLAIM=...
            # Maps to PlexUser on host with Permissions ONLY for Media share.
            - PLEX_UID=1001 
            - PLEX_GID=1000
            # Static mappings to local network.
            - TZ=...
            - HOSTNAME=...
            - ADVERTISE_IP=http://192.168.1.108:32400
            - ALLOWED_NETWORKS=192.168.1.0/24,172.16.0.0/16
        volumes:
            - config:/config
            # Map Public Media to share. QTiering is enabled to cache commonly read files.
            - /share/CACHEDEV1_DATA/Media:/share:ro
            - /etc/localtime:/etc/localtime:ro
            # Just mapping to /tmp wont work, because /tmp on QNAP is super small.
            #   - /tmp:/transcode
            # So we can either hack it, and get an autorun script to remount:
            #   https://forum.qnap.com/viewtopic.php?p=725249
            # OR just create a unique tmpfs for the container:
            # but the issue is that this directory will be mounted as root...
            #   https://github.com/docker/compose/issues/3270
            - type: tmpfs
              target: /transcode
```

In this scenario, I have to log into the terminal on every reboot/restart and perform the following:

` mkdir -p /transcode/Transcode/Sessions && chown -R plex:plex /transcode` 

This is unfortunate. I'm hoping there is an option I'm missing since I'm still ramping up on docker/compose.

PR #31, opened 2018-02-24
Fixes to version check and channel file in updater script

ianmaddox:
The update logic wasn't firing previously. It was attempting to compare the channel (beta||public) to a version number. This update fixes the faulty version check, makes the output a bit more explicit, and changes the file /version.txt to the more obvious /channel.txt, migrating any existing file.

version.txt was confusing because one would expect it to have the actual version number. In fact, it contains the update channel (public||beta). Switching file name in order to align with the verbiage inside plex.

### DIFF
--- a/root/etc/cont-init.d/50-plex-update
+++ b/root/etc/cont-init.d/50-plex-update
@@ -27,42 +27,35 @@ else
 fi
 
 # Read set version
-versionToInstall="$(cat /version.txt)"
-if [ -z "${versionToInstall}" ]; then
-  echo "No version specified in install.  Broken image"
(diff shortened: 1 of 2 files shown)


gbooker:
> The update logic wasn't firing previously.  It was attempting to compare the channel (beta||public) to a version number.

This is going to require more explanation including the exact instructions to reproduce.

> version.txt was confusing because one would expect it to have the actual version number.

And in some images it does!  Renaming this file to something else make it "confusing" in those cases.

gbooker:
This is a non-issue; code works as intended.

PR #32, opened 2018-03-09
Clarify data folder strucutre

bobbysteel:


gbooker:
This is completely wrong.  There is nothing special at all about `/data` and the container does nothing to treat it specially.  I don't have a single mount under `/data` in my containers and they work just fine.

bobbysteel:
Well I have experienced the opposite today. On a fresh image all mounts
done like /tv were mounted as d---------------. Exec'ing in as root showed
files fully viewable but plex user saw nothing. Manually changing them to
777 fixes this but implies new container will cause same issue. Why would
the default permission in container for this be set that way?

bobbysteel:
But I appreciate you channeling Linus in your polite responses :/

gbooker:
Proposing a patch with no description and only a commit message saying your are clarifying a statement when the changes to the text make the statement wrong deserves nothing short of stating that fact.  Furthermore I provided 2 independent pieces of evidence to that fact.  This changes the text to something that is wrong when the original text was correct.  That's the end of the PR.

P.S. If I were truly channeling Linus, I would have insulted your intelligence and lambasted you for even daring to propose such a change, but I did neither of those.  I fear that continuing this discussion is only going to be destructive and as such I'll not continue.

PR #34, opened 2018-07-03
Export vendor/device/model/platformver information.

chrisallen:
As per https://github.com/plexinc/plex-media-server/issues/2779 this exports better values to use for Vendor, Device, Model, and Platform Version for Plex Web to display in "devices" page.


chrisallen:
@gbooker I have fixed up the back ticks and switched to `$()`

PR #35, opened 2018-08-17
Allow overriding upstream image

acundari:
Allows changing the upstream image.

I use this to build from `nvidia/cuda:9.1-runtime` (which uses ubuntu:16.04 as its upstream) to allow GPU hardware transcoding in the container when used in conjunction with https://github.com/NVIDIA/nvidia-docker.

It'd be neat if a new tag was added for a container built with cuda support.


gbooker:
I have to question if this is really a necessity.  Typically what's done is to have the necessary drivers on the host side and passthrough the necessary devices to the container.

I did run across this page describing an alternative: https://www.funkypenguin.co.nz/note/gpu-transcoding-with-emby-plex-using-docker-nvidia/

acundari:
I couldn't get mapping /dev/dri as a device to work for the life of me, which I why I went this direction.

I didn't know about `NVIDIA_VISIBLE_DEVICES`, that's interesting. I'll test that with the public image and report back.

acundari:
So yeah, I get the same error when I try `NVIDIA_VISIBLE_DEVICES` as I was getting when I was simply mapping `/dev/dri` as a device. It appears that Plex needs the cuda driver on the container.

```
Aug 20, 2018 16:42:31.037 [0x7fdc2dbff700] ERROR - [FFMPEG] - libva: va_getDriverName() failed with unknown libva error,driver_name=(null)
Aug 20, 2018 16:42:31.037 [0x7fdc2dbff700] ERROR - [FFMPEG] - Failed to initialise VAAPI connection: -1 (unknown libva error).
Aug 20, 2018 16:42:31.037 [0x7fdc2dbff700] ERROR - [FFMPEG] - Cannot load libcuda.so.1
Aug 20, 2018 16:42:31.037 [0x7fdc2dbff700] WARN - avcodec_open2 returned -1313558101 for encoder 'h264_nvenc'
```

tzz:
> It appears that Plex needs the CUDA library on the container.

@acundari have you tried building the image yourself as you suggest and testing it? Does that work?

acundari:
@tzz Yes, I build my custom fork with `nvidia/cuda:9.1-runtime` as the base image. I have been using that for about 6 months now without issue.

ofek:
@gbooker Hello there! I also have need for this. Any chance we could get this merged soon?

gbooker:
This still strikes me as the completely wrong approach (differing base images for differing graphics cards).  Certainly it's the wrong approach to double (or more) the number of tags to differentiate those with different base images.

The link I posted above claims to work without modification of the container and is a solution posted by Nvidia themselves.  That strikes me as a better solution personally where Nvidia's docker wrapper interjects the requirements into the container.

acundari:
@gbooker The base Plex image does not have the CUDA library, so hardware transcoding cannot work with it. The instructions in your link will not work with the base image. See my reply to your comment, it's right the error is right there in the log. Reading the comments of that blog entry it seems like the solution was only tested on Emby.

gbooker:
I still don't see how changing the base image is an acceptable approach but I can see many reasons why it isn't.  Considering that's the crux of the PR, I don't why it should be accepted.

acundari:
I guess the alternative would be a second Dockerfile with the CUDA install dropped in it. Basically the steps in the image I override the base with currently (https://gitlab.com/nvidia/cuda/blob/ubuntu16.04/9.1/base/Dockerfile). At that point you have to own updating the version and potentially track multiple CUDA versions.

I only submitted this to give an easy way for people to build a CUDA compatible image via docker-compose without having to keep a personal fork.

I understand not wanting to build official images with a different base image, but that's a different question than simply _allowing_ someone to build their own image with a different base image.

brandom:
@acundari [nvidia-docker](https://github.com/NVIDIA/nvidia-docker) works perfectly for me. Did you follow all of the instructions in the link @gbooker provided above? I used the install instructions from the [nvidia-docker quickstart](https://github.com/NVIDIA/nvidia-docker#quickstart).

You need the proprietary drivers and the nvidia-docker package installed on the host before launching the container with the NVIDIA_VISIBLE_DEVICES=all environment variable. You can verify the GPU is available by running nvidia-smi in the Plex container.

gbooker:
Thanks for the comment @brandom.  I figured it had to work because NVIDIA themselves made and recommend this technique for exposing cuda/gpu to the docker container.

acundari:
@brandom I just tested and it appears that it works. I'm not sure why, but previously nvidia-docker wasn't mapping the CUDA libraries into the container for me (hence the desire to override with a base that includes them). I didn't realize it was supposed to do that. Closing this.

PR #38, opened 2019-01-02
Update s6-overlay and Ubuntu 18.04

enoch85:
As the title says.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
-ARG S6_OVERLAY_VERSION=v1.17.2.0
+ARG S6_OVERLAY_VERSION=1.21.4.0
+ARG S6_OVERLAY_MD5HASH=3eb36dc6524522d8c637106ce74ded18
 ARG DEBIAN_FRONTEND="noninteractive"
 ENV TERM="xterm" LANG="C.UTF-8" LC_ALL="C.UTF-8"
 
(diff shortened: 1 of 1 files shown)


gbooker:
Is there a reason to desire this other than just updating the the "latest and greatest" version?  The 16.04 image is still supported by Canonical.

enoch85:
>  "latest and greatest"

No, no special reason. Just what you said.

But at the same time I've seen that users had issues with the older S6 so maybe this would solve those issues (the current S6 is very old!), and I don't see any reason to stay on an old version anyway.

gbooker:
Upgrading has a risk of breakage in a scenario where it worked previously.  This is one of the reasons I've not changed this to an 18.04 base already (same with upgrading S6).  Essentially a "if it ain't broke, don't fix it" line of thinking.

> I've seen that users had issues with the older S6

Could you expound on this some more?

enoch85:
> Could you expound on this some more?

[Here](https://github.com/just-containers/s6-overlay/issues/158#issuecomment-252758867) is one example of using a later version of S6. And [here](https://github.com/just-containers/s6-overlay/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+s6) are all open issues regarding S6.

>  ain't broke, don't fix it" line of thinking.

Yeah, I've heard supervisors at several offices I've work at think like that. They run some Ubuntu 10.04 servers due to that way of thinking. :see_no_evil:

In the same way as I understand it, in my role as Sysadmin I like to work with more recent stuff and it's my job to keep it up to date. This is my effort to accomplish that.

I've looked at the Linuxserver Plex docker and they run 18.04 with similar code base, so the only thing here AFAIK is improvements, mostly in S6 as it got [a few security tweaks](https://github.com/just-containers/s6-overlay/compare/v1.17.2.0...master) as well.

TBH I haven't tested the whole Docker build process, but if you tell me how, I'd be happy to. I'm not that familiar with Docker, except running them.

Thanks!

enoch85:
OK, so I figured out how to build this, and it works as expected.

[Here](https://0bin.net/paste/aAcEuXlyWT0mBcHR#a2hSr2aCV+Yat6PNo+wqply36F882pJjuRlPH+s+rQQ) is the output.

I then run it, and connected to it. No issues AFAIK.

IMHO this can safley be merged.

gbooker:
Your comparison to those running 10.04 is a straw-man argument.  After that I'd have a hard time taking anything else you have to say seriously.

Last April, I figured I would upgrade this image to 18.04 when it had been out around a year.  Obviously this would also require checking the list everything installed in the image and removing those no longer necessary as well as ensuring all that is required is still present.  Then would come rounds of testing.  So it's not a trivial task.

P.S.  I found it amusing you say "this can safley be merged" and then commit a fix 5 minutes later.

enoch85:
> P.S. I found it amusing you say "this can safley be merged" and then commit a fix 5 minutes later.

Lol, yeah, that was quite funny actually. :)

> Your comparison to those running 10.04 is a straw-man argument.

Agreed. 

Anyway, do as you please. Close this if you don't want to merge it. Or test it properly and merge it when you feel it's safe to merge.

Happy new year btw!

PR #39, opened 2019-02-06
Speed up health check

jsclayton:
I have this in my compose file as a fix for very slow startups of the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,4 +54,4 @@ RUN \
 # Save version and install
     /installBinary.sh
 
-HEALTHCHECK --interval=200s --timeout=100s CMD /healthcheck.sh || exit 1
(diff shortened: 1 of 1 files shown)


jsclayton:
Found the docs where I based this change off long ago:

> The health check will first run interval seconds after the container is started, and then again interval seconds after each previous check completes.

So when `interval` is 200s Docker won't even try to consider it started for more than 3 minutes!

larstobi:
@gbooker This change has been published on docker hub for the :latest tag but not for the :beta tag. Could we get this pushed for :beta as well?

gbooker:
That's an issue being tracked by our build team and they will change the job to update the images.

larstobi:
@gbooker They still haven't updated the job to include this health check speed-up. The beta docker image is still 9 months old.

Could we get an update top this, so that it does not take at least 3-4 minutes before the load balancer will send traffic to Plex? Restarts are slow without this fix.

PR #41, opened 2019-03-03
Plex Media Server hardware transcoding update

madcatsu:
This PR proposes an update based on Linuxserver.io implementation to access the Intel Quick Sync kernel driver and device.

Proofs:

image

image

Would the Plex team be willing to merge and adopt this for future support?

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ A list of current and previous Intel CPU's supporting Quick Sync is available on
 
 Hardware transcoding is a Plex Pass feature that can be added to your Docker container by bind mounting the relevant kernel device to the container. To confirm your host kernel supports the Intel Quick Sync feature, the following command can be executed on the host:
 
-`sudo lspci -v -s $(lspci | grep VGA | cut -d" " -f 1)`
+`lspci -v -s $(lspci | grep VGA | cut -d" " -f 1)`
 
 which should output `Kernel driver in use: i915` if Quick Sync is available. To pass the kernel device through to the container, add the device parameter like so:
 
@@ -181,7 +181,7 @@ docker run \
 -v <path/to/plex/database>:/config \
 -v <path/to/transcode/temp>:/transcode \
 -v <path/to/media>:/data \
---device=/dev/dri:/dev/dri
+--device=/dev/dri:/dev/dri \
 plexinc/pms-docker
 ```
 
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ As a workaround you can add `- /run` to volumes in your docker-compose.yml or `-
 ## Intel Quick Sync Hardware Transcoding Support
 If your Docker host has access to a supported CPU with the Intel Quick Sync feature set and you are a current Plex Pass subscriber, you can enable hardware transcoding within your Plex Docker container.
 
-A list of current and previous Intel CPU's supporting Quick Sync is available on the Intel [website].(https://ark.intel.com/content/www/us/en/ark/search/featurefilter.html?productType=873&0_QuickSyncVideo=True) 
+A list of current and previous Intel CPU's supporting Quick Sync is available on the Intel [website](https://ark.intel.com/content/www/us/en/ark/search/featurefilter.html?productType=873&0_QuickSyncVideo=True).
 
 Hardware transcoding is a Plex Pass feature that can be added to your Docker container by bind mounting the relevant kernel device to the container. To confirm your host kernel supports the Intel Quick Sync feature, the following command can be executed on the host:
 
--- a/README.md
+++ b/README.md
@@ -160,6 +160,43 @@ plex    | s6-supervise avahi: warning: unable to spawn ./run - waiting 10 second
 ```
 As a workaround you can add `- /run` to volumes in your docker-compose.yml or `-v /run` to the docker create command.
 
+## Intel Quick Sync Hardware Transcoding Support
+If your Docker host has access to a supported CPU with the Intel Quick Sync feature set and you are a current Plex Pass subscriber, you can enable hardware transcoding within your Plex Docker container.
+
+A list of current and previous Intel CPU's supporting Quick Sync is available on the Intel [website].(https://ark.intel.com/content/www/us/en/ark/search/featurefilter.html?productType=873&0_QuickSyncVideo=True) 
+
+Hardware transcoding is a Plex Pass feature that can be added to your Docker container by bind mounting the relevant kernel device to the container. To confirm your host kernel supports the Intel Quick Sync feature, the following command can be executed on the host:
+
+`sudo lspci -v -s $(lspci | grep VGA | cut -d" " -f 1)`
+
+which should output `Kernel driver in use: i915` if Quick Sync is available. To pass the kernel device through to the container, add the device parameter like so:
+
+```
+docker run \
+-d \
+--name plex \
+--network=host \
+-e TZ="<timezone>" \
+-e PLEX_CLAIM="<claimToken>" \
+-v <path/to/plex/database>:/config \
+-v <path/to/transcode/temp>:/transcode \
+-v <path/to/media>:/data \
+--device=/dev/dri:/dev/dri
+plexinc/pms-docker
+```
+
+In the example above, the `--device=/dev/dri:/dev/dri` was added to the `docker run` command to pass through the kernel device. Once the Plex Media Server container is running, the following steps will turn on the Hardware Transcoding option:
+
+1. Open the Plex Web app.
+2. Navigate to Settings > Server > Transcoder to access the server settings.
+3. Turn on Show Advanced in the upper-right corner to expose advanced settings.
+4. Turn on Use hardware acceleration when available.
+5. Click Save Changes at the bottom.
+
+**NOTE:** Intel Quick Sync support also requires newer _64-bit versions of the Ubuntu or Fedora Linux operating system_ to make use of this feature. If your Docker host also has a dedicated graphics card, the video encoding acceleration of Intel Quick Sync Video may become unavailable when the GPU is in use. _If your computer has an NVIDIA GPU_, please install the latest Latest NVIDIA drivers for Linux to make sure that Plex can use your NVIDIA graphics card for video encoding (only) when Intel Quick Sync Video becomes unavailable._
+
+Your mileage may vary when enabling hardware transcoding as newer generations of Intel CPU's provide transcoding of higher resolution video and newer codecs. There is a useful Wikipedia page [here](https://en.wikipedia.org/wiki/Intel_Quick_Sync_Video#Hardware_decoding_and_encoding) which provides a handy matrix for each CPU generation's support of on-chip video decoding.
+
 ## Windows (Not Recommended)
 
 Docker on Windows works differently than it does on Linux; it uses a VM to run a stripped-down Linux and then runs docker within that.  The volume mounts are exposed to the docker in this VM via SMB mounts.  While this is fine for media, it is unacceptable for the `/config` directory because SMB does not support file locking.  This **will** eventually corrupt your database which can lead to slow behavior and crashes.  If you must run in docker on Windows, you should put the `/config` directory mount inside the VM and not on the Windows host.  It's worth noting that this warning also extends to other containers which use SQLite databases.
--- a/root/etc/cont-init.d/45-plex-hw-transcode
+++ b/root/etc/cont-init.d/45-plex-hw-transcode
@@ -6,7 +6,7 @@ if [ ! -e /dev/dri ]; then
 fi
 
 # Get the group IDs for the dri devices and the video group
-DEVICE_GID=$(stat -c '%g' /dev/dri/* | head -n 1)
+DEVICE_GID=$(stat -c '%g' /dev/dri/* | grep -v '^0$' | head -n 1)
 VIDEO_GID=$(getent group video | awk -F: '{print $3}')
 
 # If the video group's ID matches the group ID of the device, exit as permissions are already setup.
--- a/root/etc/cont-init.d/45-plex-hw-transcode
+++ b/root/etc/cont-init.d/45-plex-hw-transcode
@@ -1,29 +1,26 @@
 #!/usr/bin/with-contenv bash
 
-# Check for the existence of the Intel video device and stash the GID of 
-# the device, otherwise exit as the device was not mapped to the container
-if [ -e /dev/dri ]; then
-	VIDEO_GID=$(stat -c '%g' /dev/dri/* | head -n 1)
-else
+# Check to make sure the device exists.  If it doesn't exit as there is nothing for us to do
+if [ ! -e /dev/dri ]; then
 	exit 0
 fi
 
-# Check if the device GID matches the current plex user, otherwise exit as 
-# we have previously mapped this to the Plex user group
-PLEX_GID=$(getent group plex | awk -F: '{print $3}')
-if [ "${PLEX_GID}" == "${VIDEO_GID}" ]; then
+# Get the group IDs for the dri devices and the video group
+DEVICE_GID=$(stat -c '%g' /dev/dri/* | head -n 1)
+VIDEO_GID=$(getent group video | awk -F: '{print $3}')
+
+# If the video group's ID matches the group ID of the device, exit as permissions are already setup.
+if [ "${DEVICE_GID}" == "${VIDEO_GID}" ]; then
 	exit 0
 fi
 
-# Check if the GID is taken and swap to an arbitrary high GID to 
-# nominally prevent GID collisions and add the Plex user and device
-# GID to the video usergroup
-CURRENT=$(getent group ${VIDEO_GID} | awk -F: '{print $1}')
-if [ -z "${CURRENT}" ] || [ "${CURRENT}" == 'video' ]; then
-	groupmod -g ${VIDEO_GID} video
-	usermod -a -G video plex
+# Get the current group name for the device's group ID
+CURRENT_GROUP=$(getent group ${DEVICE_GID} | awk -F: '{print $1}')
+
+if [ -z "${CURRENT_GROUP}" ]; then
+	# If there is no group name for the device's group ID, change the video group to that ID.
+	groupmod -g ${DEVICE_GID} video
 else
-	groupmod -g 65533 ${CURRENT}
-	groupmod -g ${VIDEO_GID} video
-	usermod -a -G video plex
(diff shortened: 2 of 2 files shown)


kphonik:
I'm currently hacking this in with a modified `/etc/services.d/plex/run` script - definitely would be grateful for an integrated solution!

root/etc/cont-init.d/45-plex-hw-transcode
```diff
@@ -0,0 +1,25 @@
...
+	exit 0
+fi
+
+# Check if this GID matches the current plex user
+PLEX_GID=$(getent group plex | awk -F: '{print $3}')
+if [ "${PLEX_GID}" == "${VIDEO_GID}" ]; then
+	exit 0
+fi
+
+# Check if the GID is taken and swap to 65533
```
gbooker: Before the rest of my comments, I have to ask: Why not just simply add the plex user to the GID of the /dev/dri/* device instead of creating groups and/or moving them around?  I cannot find documentation on whether `usermod -G` accepts gids but `-g` does so it likely does as well.  I did see that `-g` requires the group to exist so `-G` may as well.

Was 65533 arbitrarily chosen?  Either way the comment should describe why this value.

I had to look at the following code several times to understand it so I'll ask for more commenting.

root/etc/cont-init.d/45-plex-hw-transcode
```diff
@@ -0,0 +1,25 @@
...
+CURRENT=$(getent group ${VIDEO_GID} | awk -F: '{print $1}')
+if [ -z "${CURRENT}" ] || [ "${CURRENT}" == 'video' ]; then
+	groupmod -g ${VIDEO_GID} video
```
gbooker: Could use a comment here describing something to the effect of `# Change/create the video group to match the GID of /dev/dri`

root/etc/cont-init.d/45-plex-hw-transcode
```diff
@@ -0,0 +1,25 @@
...
+	usermod -a -G video plex
+else
+	groupmod -g 65533 ${CURRENT}
```
gbooker: A comment here describing something like `# Move the current group to another id, 65533, change/create the video group to match the GID of /dev/dri and add the plex user to that group`

README.md
```diff
@@ -160,6 +160,43 @@ plex    | s6-supervise avahi: warning: unable to spawn ./run - waiting 10 second
...
 As a workaround you can add `- /run` to volumes in your docker-compose.yml or `-v /run` to the docker create command.
 
+## Intel Quick Sync Hardware Transcoding Support
+If your Docker host has access to a supported CPU with the Intel Quick Sync feature set and you are a current Plex Pass subscriber, you can enable hardware transcoding within your Plex Docker container.
+
+A list of current and previous Intel CPU's supporting Quick Sync is available on the Intel [website](https://ark.intel.com/content/www/us/en/ark/search/featurefilter.html?productType=873&0_QuickSyncVideo=True).
+
+Hardware transcoding is a Plex Pass feature that can be added to your Docker container by bind mounting the relevant kernel device to the container. To confirm your host kernel supports the Intel Quick Sync feature, the following command can be executed on the host:
+
+`sudo lspci -v -s $(lspci | grep VGA | cut -d" " -f 1)`
```
gbooker: This command on my Ubuntu 18.04 machine did not require a `sudo`.  Does it in general?

README.md
```diff
@@ -160,6 +160,43 @@ plex    | s6-supervise avahi: warning: unable to spawn ./run - waiting 10 second
...
+docker run \
+-d \
+--name plex \
+--network=host \
+-e TZ="<timezone>" \
+-e PLEX_CLAIM="<claimToken>" \
+-v <path/to/plex/database>:/config \
+-v <path/to/transcode/temp>:/transcode \
+-v <path/to/media>:/data \
+--device=/dev/dri:/dev/dri
```
gbooker: Missing the ` \` at the end of the line

README.md
madcatsu: Resolved in 8dcd1e157b27ae74f88d66e8d4422fdb871852e3

README.md
madcatsu: Resolved in 8dcd1e157b27ae74f88d66e8d4422fdb871852e3

root/etc/cont-init.d/45-plex-hw-transcode
madcatsu: @gbooker the GID of the device gets reset to root on reboot of the host, so this is a more persistent way of resolving access to the device on container first deploy/restart. So the script fetches the current GID of the device and adds that into the `video` group, and then adds the `plex` user into the same group membership.

I'll add more comments to the functions as I understand them in the script, which I've ported from the original Linuxserver.io container that uses `abc` as the base user. I'll push another commit here for you to hopefully approve 👍

root/etc/cont-init.d/45-plex-hw-transcode
madcatsu: Also, there's probably a more elegant way to grab a free GID but my bash-fu isn't that strong yet 😆

root/etc/cont-init.d/45-plex-hw-transcode
madcatsu: Should be resolved by d3ba127

gbooker:
In the case where there is a group for the GID of the the `/dev/dri/*` and it is not the `video` group, I'm not sure these three lines are really the best way to handle this.

They change the GID of that group, creates a `video` group with that GID (or change it to that GID), and then adds the `plex` user to the `video` group.  I have to wonder why not just add the `plex` user to the preexisting group?  Is there something special about it being the `video` group?

madcatsu:
`/dev/dri` is owned by 'root:video' already on the host

By default the `plex` user is already a member of the video group, and this is sufficient enough to allow hw transcoding. 

However, if the container is started with `PLEX_UID` and `PLEX_GID` values, we need to take the value of PLEX_GID (which was previously updated in `40-plex-first-run`) and ensure this group is also added to the `video` group.

Do you have a more elegant way in mind of achieving this that doesn't involve 'chowning' the device or making permissions changes on the host beforehand? Open to ideas :)

gbooker:
> By default the `plex` user is already a member of the video group, and this is sufficient enough to allow hw transcoding.

Appears to be a part of the dpkg install script to set this up.  This would make the `usermod -a -G video plex` calls unnecessary.

> However, if the container is started with `PLEX_UID` and `PLEX_GID` values, we need to take the value of `PLEX_GID` (which was previously updated in `40-plex-first-run`) and ensure this group is also added to the `video` group.

These don't change group membership; they change the ids.  They will change the UID of the `plex` user and the GID of the `plex` group.  After the change the `plex` user will still belong to the same groups as before.  Also I think you mistyped above in talking about adding the group to the `video` group and intended to mean adding the **user** to the `video` group.

My suspicion of the real issue is that the `video` group inside the container likely has a different GID than the `video` group outside the container.  It's the GIDs that matter across this barrier, not the names.
In my setup the `video` group IDs match inside and outside the container, but this isn't a surprise because my docker host is Ubuntu.  You can confirm if this is what's going on in your setup by running `getent group video` both inside and outside the container.

So, that leads us to two sets of code:

1. When the GID of the `/dev/dri` device does not correspond to any group or is the `video` group:
    ```
    groupmod -g ${VIDEO_GID} video
    usermod -a -G video plex
    ```
1. Otherwise (the GID of `/dev/dri` corresponds to a group and it is not the `video` group)
    ```
    groupmod -g 65533 ${CURRENT}
    groupmod -g ${VIDEO_GID} video
    usermod -a -G video plex
   ```

On these:
1. Not really necessary if the GID corresponds to the `video` group and the `plex` user already belongs to the `video` group.
1. Here why not just do a `usermod -a -G ${CURRENT} plex` to add the `plex` user to that group and leave the `video` group untouched?  What you are doing is changing the GID of the `$CURRENT` group to some value and then changing the GID of `video` to match the GID of `/dev/dri`.  But since it is the IDs of the groups that matter, your effective change is to remove the old GID of the `video` group and add the GID of 65533.  The only reason to do this that I can see is if there is something special about the `video` group **name** and not its ID.

madcatsu:
Great feedback!

And yes you are correct that the GID's are different across distro's. I believe from what I've read that CentOS ends up with a GID of 39 for the `video` group.

I think we should be left with something along the lines of:

- get the current GID from /dev/dri/*
- if the current GID matches the one assigned to `video`, great, do nothing
- if the current GID doesn't match, add the GID to the `plex` user
- if the PLEX_UID and PLEX_GID are specified, update PLEX_GID with GID of `video`

Apologies, I'm still pretty new to bash and I picked up the changes that Linuxserver container had implemented to achieve this. Do you think the above is the most logical way to achieve this then?

madcatsu:
OK, I stepped through this manually and the first part does actually make sense.

In a CentOS container, if you stat the group attribute from`/dev/dri/*` you get the GID from the host, which will not match the GID of the `video` group if your Docker host is Ubuntu.

Likewise, if you have an Ubuntu container on a CentOS Docker host, your GID's won't match, so ...

```bash
CURRENT=$(getent group ${VIDEO_GID} | awk -F: '{print $1}')
if [ -z "${CURRENT}" ] || [ "${CURRENT}" == 'video' ]; then
	groupmod -g ${VIDEO_GID} video
	usermod -a -G video plex
```

changes the GID of `video` group if the GID of /dev/dri/* does not match a current GID for `video` or if `video` already exists, ensures it has now got the correct GID. This part works because there would be no collision between the existing GID on /dev/dri and the video group

```bash
else
	groupmod -g 65533 ${CURRENT}
	groupmod -g ${VIDEO_GID} video
	usermod -a -G video plex
fi
```

stipulates that if neither of the above are true, then if the GID of /dev/dri/* is something else, move the flip the group ID of `VIDEO_GID` to something else momentarily from what it was (assuming it didn't actually match the GID of /dev/dri/*) and then flips it again to use the new GID

So the step `groupmod -g 65533 ${CURRENT}` moves the GID away from whatever value was default in the container OS when the Docker host isn't Ubuntu and back to the GID detected on /dev/dri/*

gbooker:
> I think we should be left with something along the lines of:
> 
> * get the current GID from /dev/dri/*

Yep.

> * if the current GID matches the one assigned to `video`, great, do nothing

Sounds good; don't do anything if there's nothing to do.

> * if the current GID doesn't match, add the GID to the `plex` user

Sounds good; this will grant the `plex` user access to the device, though I'd change it to what I have below:

> * if the PLEX_UID and PLEX_GID are specified, update PLEX_GID with GID of `video`

No.  The `PLEX_GID` is changing the GID of the `plex` group and it's purpose is typically to grant read access to the media.  Many people use sticky bits on groups and their media all belongs to a particular group and then set `PLEX_GID` to the ID of that group.  Personally, I avoid this hell by using ACLs but that's a different story.


I think you are better off following a modification to logic above which I've outlined below.  I don't think that the `plex` user needs to be added to the `video` group because it is already a member of it.  Can you confirm this in your environment?

To summarize, I think the best logic is:
* get the current GID from /dev/dri/*
* if the current GID matches the one assigned to `video`, great, do nothing
* else if there is a group with that GID, add that group to the `plex` user
* else change the `video` group to that GID

Note the last bullet point comes into play when the GID doesn't match the `video` group and there is no group that matches that GID.

madcatsu:
>  I don't think that the plex user needs to be added to the video group because it is already a member of it. Can you confirm this in your environment?

Yes. As you say that's handled by the Plex pkg install.

As for your proposed changes, I think this is what you mean. I've got no way to test though that this will work with any other distro than Ubuntu.

```bash
if [ -e /dev/dri ]; then
	DEVICE_GID=$(stat -c '%g' /dev/dri/* | head -n 1)
else
	exit 0
fi

VIDEO_GID=$(getent group plex | awk -F: '{print $3}')
if [ "${DEVICE_GID}" == "${VIDEO_GID}" ]; then
	exit 0
fi

CURRENT_GID=$(getent group ${DEVICE_GID} | awk -F: '{print $1}')
if [ -z "${CURRENT_GID}" ]; then
	groupmod -g ${DEVICE_GID} video
else
	APPEND_GID=$(getent group ${DEVICE_GID} | awk -F: '{print $3}')
	usermod -a -G ${APPEND_GID} plex
fi
```

gbooker:
I would change this to:
```bash
# Check to make sure the device exists.  If it doesn't exit as there is nothing for us to do
if [ ! -e /dev/dri ]; then
	exit 0
fi

# Get the group IDs for the dri devices and the video group
DEVICE_GID=$(stat -c '%g' /dev/dri/* | head -n 1)
VIDEO_GID=$(getent group video | awk -F: '{print $3}')

# If the video group's ID matches the group ID of the device, exit as permissions are already setup.
if [ "${DEVICE_GID}" == "${VIDEO_GID}" ]; then
	exit 0
fi

# Get the current group name for the device's group ID
CURRENT_GROUP=$(getent group ${DEVICE_GID} | awk -F: '{print $1}')

if [ -z "${CURRENT_GROUP}" ]; then
	# If there is no group name for the device's group ID, change the video group to that ID.
	groupmod -g ${DEVICE_GID} video
else
	# There is a group name, so add it to the plex user's list of groups
	usermod -a -G ${CURRENT_GROUP} plex
fi
```

Aside from comments, I changed `CURRENT_GID` -> `CURRENT_GROUP` because it is fetching a group name.  Additionally in the very last case, since the group name is known, just add that to the plex user.  Your code there had it fetch the group id, from the group id (seems redundant) and the `usermod` command requires a group name, not an id (I tested it in an ubuntu:16.04 container).

madcatsu:
Would be good to have someone on CentOS or OpenSUSE with Intel Quick Sync enabled CPU test this now as I don't have access to a host with either OS + relevant CPU

Feel free to pull `docker pull madcatsu/pms-docker-hw` and add the `--device=/dev/dri:/dev/dri` flag when running the container. 

Don't forget to pass through other relevant volumes/env vars!

1. Run `docker exec -it <container name> /bin/bash` once the container is running. 
2. Confirm GID of /dev/dri inside the container with `stat -c '%g' /dev/dri/* | head -n 1`
3. Confirm GID of `video` group matches /dev/dri with `getent group video | awk -F: '{print $3}'`
4. Enable hardware transcoding in Plex (requires Plex Pass subscription)
5. Test playback of a H.264 format video and confirm hardware transcoding working

gbooker:
Just a note saying that this PR is not forgotten.  With our new build system we had to create anew the jobs that create docker images.  Along the way we put in place the ability to create images from other branches in this repo so I'm looking to create an image from this PR so we can have more widespread testing.  Just waiting for pieces of the build infrastructure to be in place.

madcatsu:
Thanks @gbooker . This has been working quite well for me on Ubuntu, but without another physical server with iGPU I've been unable to confirm on other distro's

blkpanther5:
> Thanks @gbooker . This has been working quite well for me on Ubuntu, but without another physical server with iGPU I've been unable to confirm on other distro's

I'm running CentOS 7. Following your instructions, I replaced my standard Plex docker with your derived container, and hardware encoding seems to be working as expected. I have an older CPU (i7-2720QM), but it is QuickSync enabled.

FYI the output of `stat -c '%g' /dev/dri/* | head -n 1` matches, inside to outside the container. Not sure what you're asking for with line 3 though. 

Screenshot attached, showing hardware transcoding operating. You can also tell due to the early version of QuickSync doing a fairly poor transcode. Definitely looks worse than the software transcode.

PlexHWTranscode

nemchik:
You may consider the changes here https://github.com/linuxserver/docker-plex/pull/178 which prevent using `0` (root) in the case where the `/dev/dri` folder contains something like the `by-path` folder that my system has for some reason.

```
baines@yori:~$ ls -lah /dev/dri
total 0
drwxr-xr-x  3 root root       100 May 19 07:21 .
drwxr-xr-x 18 root root      3.9K May 19 07:21 ..
drwxr-xr-x  2 root root        80 May 19 07:21 by-path
crw-rw----  1 root video 226,   0 May 19 07:21 card0
crw-rw----  1 root video 226, 128 May 19 07:21 renderD128
```
```
baines@yori:~$ ls -lah /dev/dri/by-path/
total 0
drwxr-xr-x 2 root root  80 May 19 07:21 .
drwxr-xr-x 3 root root 100 May 19 07:21 ..
lrwxrwxrwx 1 root root   8 May 19 07:21 pci-0000:00:0f.0-car
d -> ../card0
lrwxrwxrwx 1 root root  13 May 19 07:21 pci-0000:00:0f.0-ren
der -> ../renderD128
```
```
baines@yori:~$ uname -a
Linux yori 4.15.0-50-generic #54-Ubuntu SMP Mon May 6 18:46:
08 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

Note: the first PR link above was incomplete, here's the second commit to fix matching 0 incorrectly https://github.com/linuxserver/docker-plex/pull/180#issuecomment-494227558

madcatsu:
Thanks @nemchik , have updated as recommended. I noted my server also has the `by-path` directory with root as the owner. Good work noticing that one!

JohnGalt1717:
Hi, I hope this hasn't been forgotten.

Will it work with nvidia as well? Would really like to get transcoding going!

Thanks!

Sparticuz:
I'm wondering if this will also fix my problem with /dev/dvb as well? I can't access my tv tuners.

madcatsu:
Great, been looking forward to seeing this added officially 👍

JohnGalt1717:
This is excellent! Any chance of getting the other half integrated or providing a script for what needs to happen in our own docker file that builds on this to do so?

I can't figure out what packages are necessary for the encode script inside the docker container.

madcatsu:
No problem, go for it

Sparticuz:
Gah, I missed commenting before merge! Can you update this to do the exact same thing to /dev/dvb/*?

gbooker:
@Sparticuz I take it you mean the entire `/dev/dvb` tree, not just the contents of `/dev/dvb`, right?

Sparticuz:
@gbooker Yep. I believe the reason I can't access my TV Tuners is because of the exact same issue as why we couldn't access the HW Decoder. The Linuxserver container has /dev/dvb in that script and I can access the tuners on that, but not with this container.

madcatsu:
Sorry @Sparticuz , I don't have a DVB Tuner card so I'm unable to test or see what perms are set across distro's for these. This might be as easy as copying this init script and replacing the paths so that it's independent of the Quick Sync hardware decoding checks (I can foresee a scenario where the host might not have an Intel CPU but does have a Tuner card).

PR #42, opened 2019-03-23
updated to latest version

dolchi21:
Updated base docker image and added latest available PMS deb link.


gbooker:
First part is duplicate of #38 and second is not at all how this image is managed.

PR #45, opened 2019-08-26
Also check connected tuners for video group.

gbooker:
Discussion came up in #41.  We should check dvb devices for permissions as well.

@Sparticuz, can you check these changes to see if this works for you?  I don't have the hardware to test with.  Likely best tested without the dri devices passed into the container.


madcatsu:
@gbooker - I noted that the proposed change looks for both `/dev/dri` and `/dev/dvb` and if both _**do not**_ exist then exit. This would cause a situation where users with one device or the other not installed will lose both the DVB Tuner and h/w transcoding  features.

LSIO script handles that a bit more gracefully with an `||` and then appropriate conditions if either are true.

madcatsu:
@gbooker - you're right sorry

PR #48, opened 2020-02-01
Add arm64 support and Update Plex Download URLs to latest API

maxlabuche:
(EDIT 1: Making my PR explanations more understandable)

Hi Plex Community,

I have a suggestion for the official Plex Docker Image, which is to make it available for different platforms other than Ubuntu/amd64.

**Modification suggested by my branch:**

- ADD: Add environment variables (ARGs in Dockerfile) to be able to create Docker Images for almost any OS and architecture supported by Plex Media Server by only changing these environment variables ;
- ADD: Add a Dockerfile for ubuntu/arm64 platforms ;
- EDIT: Update S6-Overlay to a more recent version, which now supports new architectures like ARM ;
- EDIT: Update package URLs to the latest Plex Download API
> PLEX_BUILD and PLEX_DISTRO environment variables correspond to the "build" and "distro" variables listed in the API 5 JSON, provided by Plex, referencing all available packages : 
> [https://plex.tv/pms/downloads/5.json](https://plex.tv/pms/downloads/5.json)

**Testing environment:**

I have tested this with a **Raspberry Pi 4** with this configuration:

Element | Version
-------- | --------
Motherboard | **Raspberry Pi 4 Model B (4GB RAM)**
Host OS | **Ubuntu Server 19.10 (eoan) [arch=arm64]**
Docker | **Docker CE for Ubuntu 19.04 (disco) [arch=arm64]**
Docker container: Plex Build | **Ubuntu 16.04+ / linux-aarch64 / debian**
Docker container: Plex pkg | **plexmediaserver_1.18.5.2309-f5213a238_arm64.deb**
  
Everything works fine, even updates.  
Please consider merging my branch to this repo, I am available for any feedback!

Greetings

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ VOLUME /config /transcode
 ENV CHANGE_CONFIG_DIR_OWNERSHIP="true" \
     HOME="/config"
 
-ARG TAG=public
+ARG TAG=beta
 ARG URL=
 
 COPY root/ /
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,6 @@ RUN \
       xmlstarlet \
       uuid-runtime \
       unrar \
-      udev \
     && \
 
 # Fetch and extract S6 overlay
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,9 @@
 FROM ubuntu:16.04
 
-ARG S6_OVERLAY_VERSION=v1.17.2.0
+ARG S6_OVERLAY_VERSION=v1.22.1.0
+ARG S6_OVERLAY_ARCH=amd64
+ARG PLEX_BUILD=linux-x86_64
+ARG PLEX_DISTRO=debian
 ARG DEBIAN_FRONTEND="noninteractive"
 ENV TERM="xterm" LANG="C.UTF-8" LC_ALL="C.UTF-8"
 
@@ -15,11 +18,12 @@ RUN \
       xmlstarlet \
       uuid-runtime \
       unrar \
+      udev \
     && \
 
 # Fetch and extract S6 overlay
-    curl -J -L -o /tmp/s6-overlay-amd64.tar.gz https://github.com/just-containers/s6-overlay/releases/download/${S6_OVERLAY_VERSION}/s6-overlay-amd64.tar.gz && \
-    tar xzf /tmp/s6-overlay-amd64.tar.gz -C / && \
+    curl -J -L -o /tmp/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz https://github.com/just-containers/s6-overlay/releases/download/${S6_OVERLAY_VERSION}/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz && \
+    tar xzf /tmp/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz -C / && \
 
 # Add user
     useradd -U -d /config -s /bin/false plex && \
@@ -45,7 +49,7 @@ VOLUME /config /transcode
 ENV CHANGE_CONFIG_DIR_OWNERSHIP="true" \
     HOME="/config"
 
-ARG TAG=beta
+ARG TAG=public
 ARG URL=
 
 COPY root/ /
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -49,7 +49,7 @@ VOLUME /config /transcode
 ENV CHANGE_CONFIG_DIR_OWNERSHIP="true" \
     HOME="/config"
 
-ARG TAG=public
+ARG TAG=beta
 ARG URL=
 
 COPY root/ /
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -18,7 +18,6 @@ RUN \
       xmlstarlet \
       uuid-runtime \
       unrar \
-      udev \
     && \
 
 # Fetch and extract S6 overlay
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,0 +1,61 @@
+FROM ubuntu:16.04
+
+ARG S6_OVERLAY_VERSION=v1.22.1.0
+ARG S6_OVERLAY_ARCH=aarch64
+ARG PLEX_BUILD=linux-aarch64
+ARG PLEX_DISTRO=debian
+ARG DEBIAN_FRONTEND="noninteractive"
+ENV TERM="xterm" LANG="C.UTF-8" LC_ALL="C.UTF-8"
+
+ENTRYPOINT ["/init"]
+
+RUN \
+# Update and get dependencies
+    apt-get update && \
+    apt-get install -y \
+      tzdata \
+      curl \
+      xmlstarlet \
+      uuid-runtime \
+      unrar \
+      udev \
+    && \
+
+# Fetch and extract S6 overlay
+    curl -J -L -o /tmp/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz https://github.com/just-containers/s6-overlay/releases/download/${S6_OVERLAY_VERSION}/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz && \
+    tar xzf /tmp/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz -C / && \
+
+# Add user
+    useradd -U -d /config -s /bin/false plex && \
+    usermod -G users plex && \
+
+# Setup directories
+    mkdir -p \
+      /config \
+      /transcode \
+      /data \
+    && \
+
+# Cleanup
+    apt-get -y autoremove && \
+    apt-get -y clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    rm -rf /tmp/* && \
+    rm -rf /var/tmp/*
+
+EXPOSE 32400/tcp 3005/tcp 8324/tcp 32469/tcp 1900/udp 32410/udp 32412/udp 32413/udp 32414/udp
+VOLUME /config /transcode
+
+ENV CHANGE_CONFIG_DIR_OWNERSHIP="true" \
+    HOME="/config"
+
+ARG TAG=public
+ARG URL=
+
+COPY root/ /
+
+RUN \
+# Save version and install
+    /installBinary.sh
+
+HEALTHCHECK --interval=5s --timeout=2s --retries=20 CMD /healthcheck.sh || exit 1
--- a/root/installBinary.sh
+++ b/root/installBinary.sh
@@ -2,6 +2,7 @@
 
 . /plex-common.sh
 
+## TODO : Make a file with these 3 variables ##
 echo "${TAG}" > /version.txt
 echo "${PLEX_BUILD}" > /plex-build.txt
 echo "${PLEX_DISTRO}" > /plex-distro.txt
--- a/root/installBinary.sh
+++ b/root/installBinary.sh
@@ -3,6 +3,8 @@
 . /plex-common.sh
 
 echo "${TAG}" > /version.txt
+echo "${PLEX_BUILD}" > /plex-build.txt
+echo "${PLEX_DISTRO}" > /plex-distro.txt
 if [ ! -z "${URL}" ]; then
   echo "Attempting to install from URL: ${URL}"
   installFromRawUrl "${URL}"
--- a/root/installBinary.sh
+++ b/root/installBinary.sh
@@ -2,10 +2,10 @@
 
 . /plex-common.sh
 
-## TODO : Make a file with these 3 variables ##
-echo "${TAG}" > /version.txt
-echo "${PLEX_BUILD}" > /plex-build.txt
-echo "${PLEX_DISTRO}" > /plex-distro.txt
+addVarToConf "version" "${TAG}"
+addVarToConf "plex_build" "${PLEX_BUILD}"
+addVarToConf "plex_distro" "${PLEX_DISTRO}"
+
 if [ ! -z "${URL}" ]; then
   echo "Attempting to install from URL: ${URL}"
   installFromRawUrl "${URL}"
--- a/root/plex-common.sh
+++ b/root/plex-common.sh
@@ -19,7 +19,9 @@ function getVersionInfo {
     channel=8
   fi
   
-  local url="https://plex.tv/downloads/details/1?build=linux-ubuntu-x86_64&channel=${channel}&distro=ubuntu"
+  local plexBuild="$(cat /plex-build.txt)"
+  local plexDistro="$(cat /plex-distro.txt)"
+  local url="https://plex.tv/downloads/details/5?build=${plexBuild}&channel=${channel}&distro=${plexDistro}"
   if [ ${tokenNeeded} -gt 0 ]; then
     url="${url}&X-Plex-Token=${token}"
   fi
--- a/root/plex-common.sh
+++ b/root/plex-common.sh
@@ -19,6 +19,7 @@ function getVersionInfo {
     channel=8
   fi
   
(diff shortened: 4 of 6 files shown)


Dockerfile
```diff
@@ -45,7 +49,7 @@ VOLUME /config /transcode
 ENV CHANGE_CONFIG_DIR_OWNERSHIP="true" \
     HOME="/config"
 
-ARG TAG=beta
+ARG TAG=public
```
gbooker: Any reason for this change?

root/installBinary.sh
```diff
@@ -3,6 +3,8 @@
 . /plex-common.sh
 
 echo "${TAG}" > /version.txt
+echo "${PLEX_BUILD}" > /plex-build.txt
+echo "${PLEX_DISTRO}" > /plex-distro.txt
```
gbooker: I'm starting to wonder if this should instead be a single file that contains three variables.  When there was only one file, it was fine to have only that single variable but now that there are 3, maybe a single file containing the three variables would make more sense.  To make such a change, it looks like https://github.com/plexinc/pms-docker/blob/master/root/etc/cont-init.d/50-plex-update would also need some updating but that would be it.

root/plex-common.sh
```diff
@@ -19,7 +19,9 @@ function getVersionInfo {
     channel=8
   fi
   
-  local url="https://plex.tv/downloads/details/1?build=linux-ubuntu-x86_64&channel=${channel}&distro=ubuntu"
+  local plexBuild="$(cat /plex-build.txt)"
+  local plexDistro="$(cat /plex-distro.txt)"
```
gbooker: Would need similar changes if we have a single file.

Dockerfile.arm64
```diff
@@ -0,0 +1,61 @@
...
+ARG TAG=public
+ARG URL=
+
+COPY root/ /
+
+RUN \
+# Save version and install
+    /installBinary.sh
+
+HEALTHCHECK --interval=5s --timeout=2s --retries=20 CMD /healthcheck.sh || exit 1
```
gbooker: Can't say I like the idea of having two nearly identical files.  It'd be too easy to update one and forget to update the other.  Though it seems that the way docker is designed, the only solution to fix this is to have a template and then a script which generates the individual dockerfiles.

Dockerfile
maxlabuche: Indeed, there isn't. Actually the Download API doesn't really care about this tag, it returns the same file for both Beta and Public. If I suggested to use Public was in case a new Beta package was released and suddenly makes Plex not working on Docker. But that should not happen, so we can stay with Beta.

root/installBinary.sh
maxlabuche: Yes, you are definitely right. I can try to merge these three variables into one file, then adapt the following scripts: `installBinary.sh`, `plex-common.sh`, `50-plex-update`

root/plex-common.sh
maxlabuche: Agree. 👍

Dockerfile.arm64
maxlabuche: I have taken this idea by browsing different repo about Docker images. And yes, because Docker needs separated Dockerfile for each CPU architecture it's easy to forget one to update. I can also work around this idea of templating the Dockerfile!

maxlabuche:
> Some comments to facilitate discussion.

Thank you very much @gbooker for your feedback, I will work around all of the discussed points, then submit the following modifications during the next days! 👍

Dockerfile.arm64
gbooker: This was more of a general comment about the way Docker is setup.  Don't see this as a hard requirement for this PR.

maxlabuche: I get the idea, so what do you think, should we go for a combo template/script or stay with the 2 Dockerfile for that moment?

gbooker: For now likely the 2 Dockerfiles would be better.

root/plex-common.sh
```diff
@@ -1,5 +1,25 @@
...
+function readVarFromConf {
+  local variable="$1"
+  declare -n value=$2
+  if [ ! -z "${variable}" ]; then
+    value="$(grep -w ${variable} ${CONT_CONF_FILE} | cut -d'=' -f2 | tail -n 1)"
+  else
+    value=NULL
+  fi
+}
+
```
maxlabuche: @gbooker I think there are better ways to write different variables from a "conf" file, but I wanted to keep things simple given that these variables are written once in this file (while building the image), then they are read only.

Dockerfile.arm64
maxlabuche: Okay, that seems legit

maxlabuche:
@gbooker I have made the changes we discussed last week to the corresponding files. I am waiting for your feedback, do not hesitate to suggest any enhancement you're thinking they are relevant :)

root/plex-common.sh
```diff
@@ -1,5 +1,25 @@
...
+  if [ ! -z "${variable}" ]; then
+    echo ${variable}=${value} >> ${CONT_CONF_FILE}
+  fi
+}
+
+function readVarFromConf {
+  local variable="$1"
+  declare -n value=$2
+  if [ ! -z "${variable}" ]; then
+    value="$(grep -w ${variable} ${CONT_CONF_FILE} | cut -d'=' -f2 | tail -n 1)"
```
maxlabuche: ## Explanations for reading value :

Expecting file should be like:
```
version=beta
plex-distro=debian
...
```

Value is : look for variable name (grep), then take the part after the `'='` sign (cut), then take the last value if multiple (tail) (that should not happen but in case).

root/plex-common.sh
gbooker: Since the values are written once and in only one place, the function to add them isn't really needed (doesn't hurt though).  I was expecting just a
```
echo "version=${TAG}" > /version.txt
echo "plex_build=${PLEX_BUILD}" >> /version.txt
echo "plex_distro=${PLEX_DISTRO}" >> /version.txt
```
but the above should work too.

gbooker:
Overall this is looking good.  I need to have a discussion with our build-team to see if we can make our CI build the ARM64 images as well.

root/plex-common.sh
maxlabuche: Yes exactly I was thinking to write the same lines at the beginning like you are suggesting. At least if at some point we need to have such as a function, we just have to modify the behaviour of this one

maxlabuche:
> Overall this is looking good. I need to have a discussion with our build-team to see if we can make our CI build the ARM64 images as well.

Ok nice, I am looking forward to see what's happening next, please let me know :)
Thank you for your review and your time!

half2me:
@maxlabuche @gbooker I would really be interested in this, as I want to have a setup with the raspberry pi 4. Let me know if there is anything I can help with.

maxlabuche:
> @maxlabuche @gbooker I would really be interested in this, as I want to have a setup with the raspberry pi 4. Let me know if there is anything I can help with.

Hi @half2me , 

I am glad to see other people interested in supporting ARM architectures :)
Indeed, I have worked on this and submitted it to plexinc, but as you might see it's still in review process by the CI team (integration team).

I have made some tests at home and it's working fine for me (with RPi 4 / 4GB, better with at least 2GB of RAM on arm64), so if you are interested in testing it, you can pull my image on my Docker registry before the integration in the official repository.
 
(Instructions are exactly the same as the official one, and update checking is automatic each time you start the container)
  
https://hub.docker.com/r/maxlbch/pms-docker-arm64  
  
Enjoy!

half2me:
@maxlabuche I have the rpi4 4gb as well, so I will try your image. Thanks :)

maxlabuche:
You're welcome :)

half2me:
> You're welcome :)

@maxlabuche any chance you could build your image against amd64 arch and upload? I'm trying to make my own helm chart for plex media server, and I would like to use your modifications on arm64 and amd64 as well.

maxlabuche:
@half2me Yes that was my base idea to have multiple arch under the same tag, I just forgot to do that, I am working on it I'll let you know when it's pushed to the registry :)

half2me:
@maxlabuche a bit off topic, but if its of any help, I use github actions to build my docker images automatically on each push using `docker buildx`. You can build for any arch you want. Here is an example: https://github.com/half2me/charts/blob/master/.github/workflows/dockerimage.yml

maxlabuche:
@half2me Oh that's this kind of stuff I am looking for, indeed I was using buildx to build my images but for some reasons it failed for arm64, that's why I made it directly from the Raspberry, I will look around, thanks!

maxlabuche:
I remember, here the thing is that the Dockerfile is not the same for amd64 and arm64, that's what I should consider

half2me:
> I remember, here the thing is that the Dockerfile is not the same for amd64 and arm64, that's what I should consider

Do you think it would be worth making it the same, and branching via an ARG? If most of it is similar...

maxlabuche:
We need to look around before, for instance here is the muti-arch image :  
https://hub.docker.com/r/maxlbch/pms-docker  
Hope that it correspond in what you're looking for :)

half2me:
> We need to look around before, for instance here is the muti-arch image :
> [hub.docker.com/r/maxlbch/pms-docker](https://hub.docker.com/r/maxlbch/pms-docker)
> Hope that it correspond in what you're looking for :)
> 
> I will be working on your suggestion to be even more futureproof.

Exactly what I was looking for, thank you!
Let me know if you need any help, (dm me on twitter or keybase)

garga3838:
docker run \
-d \
--name plex \
--network=host \
-e TZ="<timezone>" \
-e PLEX_CLAIM="<claimToken>" \
-v <path/to/plex/database>:/config \
-v <path/to/transcode/temp>:/transcode \
-v <path/to/media>:/data \
plexinc/pms-docker

ipeacocks:
Why this can't be merged?

Dockerfile
```diff
@@ -15,11 +18,12 @@ RUN \
       xmlstarlet \
       uuid-runtime \
       unrar \
```
ipeacocks: Do you really need `unrar` and all of these programs?

maxlabuche: Hi @ipeacocks, actually I am not an expert of the Plex Core and features, so I can't answer you about that.. Maybe @gbooker can bring you more details about that. 

I just added `udev` because can be useful to access external storage like in my Raspberry project

maxlabuche:
> Why this can't be merged?

I submitted my work, seems like CI team is still reviewing / testing it.. Do you have any news @gbooker ? Thanks, hope you guys are doing great. Stay safe

Dockerfile
ipeacocks: A bit offtop. Why do you use 64 bit OS for Raspberry Pi. Raspberry advices to use 32-bits Raspbian even on 64-bits Pi 4.

maxlabuche: I use Pi 4 (RAM 4GB) with lots of containers on it (maria-db, postgre-db, web servers, python server, plex...) All of these is consuming lots of RAM and even more when several people are connected simultaneously. I am also using Ubuntu Server for ARM (no GUI), not Raspbian.

I haven't made deep benchmarks, but I can tell there is a significant gain in global performance and responsiveness using 64-bit with such a config. 32-bit architectures have per-process limits when it comes to use large amount of RAM.

Hope that's answering your question 🙂

half2me: @ipeacocks just like @maxlabuche I also 64bit. I believe in the future, this will be the only option, its just a matter of time. I'd like to make the switch as early as possible. Plus the performance gains :)

gbooker: > Do you really need `unrar` and all of these programs?

`unrar` is needed for some of the third-party plugins (subtitles IIRC) but maybe it's not needed anymore.  The rest of the above are definitively needed.

> I just added `udev` because can be useful to access external storage like in my Raspberry project

`udev` doesn't work as you would expect in containers.  I'm skeptical of this and missed it in the first pass through.

maxlabuche: OK @gbooker for `udev`, we can skip it if not needed, I have also figured out other solutions. So what is missing to merge this PR? Do you have any concerns?

gbooker: In my experimentation, `udev` didn't work inside the containers.  We used to have it there but its lack of functionality prompted us to remove it.

We are still looking at getting our CI to build ARM images but that may take longer than expected.

dekimsey:
It might be worth renaming the Dockerfile to `.aarch64` to follow the architecture being used. The file is currently `arm64` but the code refers to `aarch64`.

maxlabuche:
@dekimsey You're right, but on Docker Hub 64-Bit ARM architectures are under the `arm64` tag and not `aarch64`, that's why I'm using this naming method

Dockerfile
maxlabuche: Thanks @gbooker , do not hesitate to add new comments.

Dockerfile
```diff
@@ -15,11 +18,12 @@ RUN \
...
+      udev \
```
gbooker, line 21: Should be removed since it doesn't work in containers.

Dockerfile.arm64
```diff
@@ -0,0 +1,61 @@
...
+RUN \
+# Update and get dependencies
+    apt-get update && \
+    apt-get install -y \
+      tzdata \
+      curl \
+      xmlstarlet \
+      uuid-runtime \
+      unrar \
+      udev \
```
gbooker, line 21: `udev` should be removed since it doesn't work in containers.

half2me:
@gbooker not sure what you use in your CI to build arm images, but I've had success using docker buildx to cross-compile for arm. If there is anything I can help with, let me know.

JayAndCatchFire:
@half2me Speaking of buildx, I'm trying to build using this setup (with my [armhf addition](https://github.com/JayAndCatchFire/pms-docker)) and buildx to make all the versions.  But I'm not seeing how to tell buildx that it's one build but different Dockerfiles based on the architecture.  Basically, it's trying to use `Dockerfile` for all the different architectures, which technically succeeds, but then all the binaries are wrong.

Any ideas on how to do what I'm describing so that pms-docker:latest would work for all supported architectures?  Or should we maybe consider unifying the Dockerfiles into a single one that builds correctly no matter what architecture you're on?

JayAndCatchFire:
FWIW, I rewrote [my version](https://github.com/JayAndCatchFire/pms-docker) to support x86, x86_64, aarch64, and armhf using a single Dockerfile so it's far more buildx friendly.

half2me:
@JayAndCatchFire personally I like having a single Dockerfile whenever possible, but if there are too many differences between archs, you are better off with multiple Dockerfiles. You can simply tell docker buildx to use a specific Dockerfile for the build. If you give them all the same tag, they should be consolidated.

maxlabuche:
> Getting our CI to build ARM64 images is going to take longer so I'm going to target merging this without that. I think that `udev` shouldn't be in the image since it doesn't actually work in containers but otherwise I'm fine with the changes in this PR.

@gbooker OK, that's done, I just removed `udev` as it seems unnecessary. I think we are good, you can merge if you do not have more requests.

@JayAndCatchFire I like your work, for now I just wanted to add minor updates to this repo, that's why I didn't merge your PR to my branch. Also well suggested by @half2me , working on a multi-arch Dockerfile is definitely a good idea. Currently with the worldwide situation I can not put a lot of time on this project but I can help you if you need support implementing that (with buildx, scripting, etc.)

Thank you all for your very interesting reviews! Do not hesitate to @ me in your comments if you need further help. Take care.

JayAndCatchFire:
@half2me Please tell me more about that.  When I ran buildx on separate Dockerfiles serially, it replaced the latest tag on Docker Hub for me rather than adding to it.

PR #49, opened 2020-04-07
Create Dockerfile.armv7

JayAndCatchFire:
Since we've got one for arm64 now, why not also `armhf`/`arm/v7`?  I'm successfully using this build on my Raspberry Pi 4 B running Raspbian.

### DIFF
--- a/Dockerfile.armv7
+++ b/Dockerfile.armv7
@@ -1,0 +1,60 @@
+FROM ubuntu:16.04
+
+ARG S6_OVERLAY_VERSION=v1.22.1.0
+ARG S6_OVERLAY_ARCH=armhf
+ARG PLEX_BUILD=linux-armv7hf_neon
+ARG PLEX_DISTRO=debian
+ARG DEBIAN_FRONTEND="noninteractive"
+ENV TERM="xterm" LANG="C.UTF-8" LC_ALL="C.UTF-8"
+
+ENTRYPOINT ["/init"]
+
+RUN \
+# Update and get dependencies
+    apt-get update && \
+    apt-get install -y \
+      tzdata \
+      curl \
+      xmlstarlet \
+      uuid-runtime \
+      unrar \
+    && \
+
+# Fetch and extract S6 overlay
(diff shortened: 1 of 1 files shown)


gbooker:
I do have to wonder about the use-case for this architecture.  With the rPi4 I could see the use for the 64-bit version because it has a good chunk of RAM and can absorb the docker overhead but I'm wondering about the usefulness on 32-bit platforms.

JayAndCatchFire:
I can't speak for the earlier models of Raspberry Pi, but I'm running Raspbian on it right now with this Plex server image and I'm having no trouble at all.

gbooker:
> I can't speak for the earlier models of Raspberry Pi

Which model are you running?

JayAndCatchFire:
>I'm successfully using this build on my Raspberry Pi 4 B running Raspbian.

gbooker:
That has a 64-bit processor and running 64-bit code would be faster.  I'm wondering the scenario where one cannot run 64-bit on arm and if the docker overhead is worth it in that case.

JayAndCatchFire:
I've personally never personally felt any docker overhead.  Especially on the speedy Pi 4.  Maybe docker images are a little slower to start or stop compared to x86_64, but once they're running, it's felt the same as running in the host OS.

It's easy enough to put the Raspberry Pi 4 B into 64-bit mode while running Raspbian, which is my preferred OS for the Pi.  However, since the userspace is still 32-bit, installing the 64-bit version of docker can't really be done without a lot of headache and heartache.  At least, I've been unable to do it.  When your userspace is 32-bit, the docker install scripts want you to use the armhf version of docker.  And from what I've read, Raspbian will be sticking to a 32-bit userspace for the foreseeable future in order to remain compatible with the entire Pi line.

If this is an issue, it's okay to reject my PR.  I can maintain my own fork of this repo for arm7l compatibility for my own use.  I just figure it's a logical, low maintenance move since s6 overlay and plex itself both support arm7l.  And I've definitely seen interest in running this on docker out there in the form of articles, blog posts, other Github repos, and Docker Hub images.

gbooker:
> Raspbian will be sticking to a 32-bit userspace for the foreseeable future in order to remain compatible with the entire Pi line.

Sounds like Windows 😉 

Sounds like you gave a use-case which is what I was looking for.

PR #50, opened 2020-05-03
Added helper script to allow periodic updates

adrianrudnik:
I'm unsure about the contribution guidelines here, so I'll just create a PR for this.

This little helper script uses the available configuration, plex claim token and available tools to check for a new version of plex. If found, it kills the primary process for that docker container. That allows for docker host systems with a restart directive to reboot and download the newest binary without any additions/modifications to that mechanic.

This allows to be configured through host systems crontab, see last section in the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,3 +206,15 @@ Docker on Windows works differently than it does on Linux; it uses a VM to run a
 If the claim token is not added during initial configuration you will need to use ssh tunneling to gain access and setup the server for first run. During first run you setup the server to make it available and configurable. However, this setup option will only be triggered if you access it over http://localhost:32400/web, it will not be triggered if you access it over http://ip_of_server:32400/web. If you are setting up PMS on a headless server, you can use a SSH tunnel to link http://localhost:32400/web (on your current computer) to http://localhost:32400/web (on the headless server running PMS):
 
 `ssh username@ip_of_server -L 32400:ip_of_server:32400 -N`
+
+## Automatic updates
+
+For periodic updates through a scheduler like contab you can use the helper script available inside the images.
+
+The following example would work on the host system with crontab:
+
+```
+0 2 * * * cd path/to/your/plex && docker-compose exec plex /updatecheck.sh
+```
+
+The script would run daily at 02:00 and force a reboot of the plex container if the local version does no longer match the available remote plex version (based on the configured claim token).
--- a/root/updatecheck.sh
+++ b/root/updatecheck.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/with-contenv bash
+
+if !(cat /proc/1/cgroup | grep -q '/docker/'); then
+  echo "Not a docker container"
+  exit 0
+fi
+
+# If we are debugging, enable trace
+if [ "${DEBUG,,}" = "true" ]; then
+  set -x
+fi
+
+. /plex-common.sh
+
+pmsApplicationSupportDir="${PLEX_MEDIA_SERVER_APPLICATION_SUPPORT_DIR:-${HOME}/Library/Application Support}"
+prefFile="${pmsApplicationSupportDir}/Plex Media Server/Preferences.xml"
+
+function getPref {
+  local key="$1"
+
+  xmlstarlet sel -T -t -m "/Preferences" -v "@${key}" -n "${prefFile}"
+}
+
+token="$(getPref "PlexOnlineToken")"
+
+# Determine current version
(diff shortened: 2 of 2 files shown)


gbooker:
TBH, I'm not sure such an operation really belongs in the container as killing the main proc seems like a poor solution.  This seems to me to better belong outside the container and execute a `docker restart …`.  For something running in the container, I'd far more expect such a script to grab and install the dpkg rather than restarting the container.

Joshfindit:
Since this is designed to be called on the host anyway, why not call `docker-compose pull` or something similar?

adrianrudnik:
@gbooker But thats exactly how the image is working right now, looking at https://github.com/plexinc/pms-docker/blob/master/root/installBinary.sh. As soon as the image boots, it installs the latest version. I do not see a reason or way to keep the container alive. As soon as the current process of plex the container would be killed anyway, so there would be no difference.

@Joshfindit As the plex binary is not part of the docker image, there is nothing to pull for plex updates. I would agree if it would be. The process itself would become much more cleaner, but as it is right now, the only way to install the latest version is to restart the docker container.

Joshfindit:
> @Joshfindit As the plex binary is not part of the docker image, there is nothing to pull for plex updates. I would agree if it would be. The process itself would become much more cleaner, but as it is right now, the only way to install the latest version is to restart the docker container.

Ahh perfect. Just `docker-compose restart` then.

adrianrudnik:
@Joshfindit Yeah, thats enough. Sadly that comes with caveat that it would always kill the server, even when no updates are available.

gbooker:
> As the plex binary is not part of the docker image, there is nothing to pull for plex updates.

There are still package updates as well as updates to the upstream Ubuntu image, so this isn't entirely true.  I am trying to get us to update and push these images more often.

Also, not everyone has `docker-compose`.

adrianrudnik:
@gbooker Sure, but docker restart policy is not part of compose, its part of docker so that would work as well. The images itself can be updated, but that has nothing to do with plex and the way plex loads the most current version from my understanding?

Joshfindit:
> @Joshfindit Yeah, thats enough. Sadly that comes with caveat that it would always kill the server, even when no updates are available.

Fair enough, but that’s what maintenance windows are for. To have the longest uptime, update as little as possible. To keep on the leading edge of updates, do a scheduled maintenance window. This is a classic server-management choice.

Joshfindit:
That being said, I personally would much rather the docker container move to including the binary so that `docker pull` *is* the right choice.
Sometimes I need to restart for some reason and *don’t* want to also update.

Then the image could also be tagged with the Plex server version.

gbooker:
> That being said, I personally would much rather the docker container move to including the binary so that `docker pull` _is_ the right choice.
> Sometimes I need to restart for some reason and _don’t_ want to also update.
> 
> Then the image could also be tagged with the Plex server version.

We have those.  `latest` will point to the latest image that was pushed which is tagged with the version of the server contained in the image.  See the list: https://hub.docker.com/r/plexinc/pms-docker/tags/  We just don't do it for beta versions but only for publicly released version.

Joshfindit:
> > That being said, I personally would much rather the docker container move to including the binary so that `docker pull` _is_ the right choice.
> > Sometimes I need to restart for some reason and _don’t_ want to also update.
> > Then the image could also be tagged with the Plex server version.
> 
> We have those. `latest` will point to the latest image that was pushed which is tagged with the version of the server contained in the image. See the list: https://hub.docker.com/r/plexinc/pms-docker/tags/ We just don't do it for beta versions but only for publicly released version.

That doesn’t quite fit as I would be running `latest` most of the time, and restarts may not be planned (a planned restart could *theoretically* look up the current version, shut down Plex, “pin” it by specifying that version as a tag in the docker-compose file, then restarting).
In my ideal scenario, Plex would simply stay at the currently-pulled `latest` version until I pull the new version.

To appease the UI-updaters in that scenario, it would be chill if updates in the Plex UI itself did update the container’s version the way it does now as it would just mean I’d *have* to pull that version or later in order to keep the DB from complaining.


Docker really works best (at least in my experience) if the containers are:
- Not holding state of any kind
- Immutable unless thoughtfully changed by the admin (like a rebuild or a pull)

gbooker:
I think you are conflating things.  `latest` is always tagged to an image that is tagged with a specific version.  If you restart one of these image, they will not change versions.  You **MUST** pull to get an update of these.  These images work exactly as you would expect from a docker container.

The only images that update the server version on restart are the `public` and `beta` tagged versions which are the only ones that could benefit from any of the prior discussion in this PR.

PR #51, opened 2020-05-20
Use a .env file

stunney:
Instead of having the user modify this boilerplate file, have them put their secrets in a .env file that is loaded by docker-compose be default.

### DIFF
--- a/docker-compose-host.yml.template
+++ b/docker-compose-host.yml.template
@@ -1,16 +1,17 @@
-version: '3.6'
+version: '3.3'
 services:
   plex:
     container_name: plex
     image: plexinc/pms-docker
+    ports:
+      - "32400"
     restart: unless-stopped
-    env_file:
-      - ./.env
     environment:
       - TZ=${TIMEZONE}
       - PLEX_CLAIM=${CLAIM_TOKEN}
+      - PLEX_HOME=${PLEX_HOME}
     network_mode: host
     volumes:
-      - <path/to/plex/database>:/config
-      - <path/to/transcode/temp>:/transcode
-      - <path/to/media>:/data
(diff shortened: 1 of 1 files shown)


docker-compose-host.yml.template
```diff
@@ -1,14 +1,17 @@
-version: '2'
+version: '3.3'
 services:
   plex:
     container_name: plex
     image: plexinc/pms-docker
+    ports:
+      - "32400"
```
gbooker, line 7: Ports are not specified in host networking mode and this is the wrong way to do so.

docker-compose-host.yml.template
```diff
@@ -1,14 +1,17 @@
...
+      - PLEX_CLAIM=${CLAIM_TOKEN}
+      - PLEX_HOME=${PLEX_HOME}
     network_mode: host
     volumes:
-      - <path/to/plex/database>:/config
-      - <path/to/transcode/temp>:/transcode
-      - <path/to/media>:/data
+      - ${PLEX_HOME}/config:/config
+      - ${PLEX_HOME}/transcode:/transcode
+      - ${PLEX_HOME}/data:/data
```
gbooker, line 17: Far too constraining to force all of these directories into the same directory tree and to force these specific directory names.

docker-compose-host.yml.template
stunney: I missed the host mode portion.  This was out of habbit.  Removing.

docker-compose-host.yml.template
stunney: You are correct, this should be changed.  I will break it out.

stunney:
The CLAIM_TOKEN is 100% a secret.  The fact that it is ephemeral doesn't make it any less so.

PR #52, opened 2020-05-21
Persistent config - for NFS-shared config directories

NateTheGreat68:
This is a seemingly-extreme but robust solution to preventing the instability or data corruption issues associated with the /config volume being located on an NFS share (possibly other filesystems that lack file-locking or have other issues).

If a user mounts the config volume on /config_persistent (instead of the usual /config) and provides a valid X-PLEX-TOKEN as environment variable X_PLEX_TOKEN, the container will automatically copy everything over from the persistent volume to /config when the container is started, then restart the plex service every day at 4am after first verifying that there are no active streams. When the service goes down, it'll copy the contents of /config to the persistent volume (in a unique subdirectory, so that there is no data loss if the copy is interrupted), then delete the old data only after the copy is completed.

The behavior can be customized using several new environment variables, described in the last section of the README.

New packages required for this all to work: rsync and cron.


NateTheGreat68:
This is available in a public dockerhub repo: natethegreat68/pms-docker:v0.10
if you want to quickly check out a built image. It has the default "beta" tag argument.

gbooker:
> instability or data corruption issues associated with the /config volume being located on an NFS share

NFS can be configured to support file locking, at least to an extent.  I believe that some have had success with it.

This strikes me as something that shouldn't be in the container but rather be completely external to it.  Essentially this is a backup mechanism.

NateTheGreat68:
> NFS can be configured to support file locking, at least to an extent. I believe that some have had success with it.

I actually use NFSv4 (with FreeNAS as the server) and still have pretty persistent performance and stability issues when the NFS share is directly mounted to /config. Letting the data exist in /config without mounting an outside volume has resolved these issues for me.

> This strikes me as something that shouldn't be in the container but rather be completely external to it. Essentially this is a backup mechanism.

I tried to go this route at first - mounting a directory on the local filesystem, and just running backups of that to the NFS share. The difficulty comes with setups using something like, in my case, RancherOS as the Docker host. Making and maintaining a local directory in RancherOS and then backing it up is messy, unless there's some functionality I'm not aware of.

Joshfindit:
I din’t know if this is a viable solution for the majority of people.

The config directory has a *frightening* amount of stuff in it. Everything from SQLite db files, to the thumbnail cache and even the `plexmediaserver.pid` file, and copying a “persistent” version may introduce all sorts of issues, from missing thumbnails to DB corruption if a Plex update is applied in between.

The other thing, as mentioned, is that this could be done outside the container using the same process of stopping the container and manipulating the hostOS-side config/config_persistent folders. Difficulties with RancherOS aside. (Personally, I’m mounting local volumes through Proxmox’s LXC containers and passing those through to Docker).
Multiple decades of filesystem experience tells me to avoid network storage layers *if at all possible*. In your case I would start exploring different alternatives to RancherOS.

bjeanes:
I also have this locking problem but I agree this solution is not a great fit. My config directory is gigabytes big with thumbnails etc and is lots of small files which are patholologically slow to copy, even with rsync, and even to local disks, let alone over the network.

Honestly, by far the best thing I've done is simply mount the .db files from the host machine (layered over the `/config` from NFS share) and relied on Plex's existing DB backup facility to write those backups to the NFS-mounted directory. The locking issue is really only ever a real problem on the databases as nothing else has a high degree of write contention. The one exception might be logs, which I also mount from a local volume.

In this approach, disaster recovery is just copying the DB backup back to the Plex server machine. This has essentially entirely fixed any noticeable perf problem.

NateTheGreat68:
Thanks for all the input everyone; I'm going to continue using and tweaking this in my own fork since it's working well for me, but I'll also be working on a better, less-ridiculous permanent fix using some of these tips. I'll also make sure to update my README file to clearly state that it's not the official image. Feel free to close this pull request out.

NateTheGreat68:
Actually, one more question: if I put together a branch that just had the portion of `root/etc/services.d/plex/finish` that's responsible for hitting the actual Plex process with a SIGTERM so that it can be gracefully stopped/restarted via s6, then added a couple scripts somewhere so that it's a lot easier to control the service from a docker exec call from outside the container, would you entertain merging that in? Or are you solidly against starting/stopping the service rather than the whole container?

gbooker:
My biggest issue here is that this solution doesn't seem like a good fit for the container.  To me this just feels like it should be external to the container because its usefulness really dependent on circumstances which are external to the container (such as the capabilities of the FS containing the database).

Personally, I ran PMS for a while on a VM where I wanted the whole config dir to be stored on redundant storage but I didn't want to give redundant storage to the VM.  I solved this (for many containers, not just Plex) by using ZFS in the VM since my redundant storage uses ZFS.  I had a cron job that would snapshot the FSs in the VM and zfs send/recv them to the redundant storage.  Since the snapshot mechanism is atomic and the database uses a write-ahead-log, the database is always consistent in the snapshots.

BTW, logs likely wouldn't have an issue on NFS because they are only appended and don't use file-locking.  Databases do use file-locking and rely on such locks to be respected across processes as well as multiple open fds within a process.  The logging mechanism only ever has a single fd open at once and is all processed in a single worker thread.

Per an S6 service, it feels to me like it should be a separate S6 service (assuming that S6 supports service ordering).  But even then it feels to me like it still doesn't belong in the container.

I'm going to close this for now but don't feel like you cannot continue discussion or re-open should circumstances change.

bjeanes:
> BTW, logs likely wouldn't have an issue on NFS because they are only appended and don't use file-locking. 

agreed. but if they have parallel writers and use locks, they could still have write contention, was my point. i have never observed this to be a performance issue and my choice to store them locally was unrelated.

PR #53, opened 2020-05-25
Stop plex service with the s6 supervisor

NateTheGreat68:
As-is: Attempting to stop or restart the plex service via s6 (with `s6-svc -d /var/run/s6/services/plex` or `s6-svc -r /var/run/s6/services/plex`, respectively) does not stop the actual Plex Media Server process as expected because it forks.

The root/etc/services/plex/finish script in this pull request sends a SIGTERM to the primary PMS process, then waits for all plex processes to terminate before the script exits. This, in combination with the basic script at root/plex_service.sh, allows end users to easily stop/restart the service from outside the container with something like `docker exec -it <container_name> /plex_service.sh -d` in order to perform maintenance such as backups.

Currently, the primary reason to stop and start the service rather than the container would be to avoid the update cycle on the beta/public images. Although it doesn't seem to matter, it also does let the Plex Media Service process terminate gracefully via receiving a SIGTERM (rather than being forcefully killed when the container is stopped, which I believe is what happens currently).


root/etc/services.d/plex/finish
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/with-contenv bash
+
+echo "Stopping Plex Media Server."
+pid_file="$(echo ~plex)/Library/Application Support/Plex Media Server/plexmediaserver.pid"
+
+# Terminate the process.
+kill -SIGTERM "$(cat "$pid_file")"
+
+# Make sure all plex processes die.
+ps aux | awk '{ if ( $0 ~ /[Pp]lex/ && $0 !~ /s6-supervise/ ) system("tail --pid="$2" -f /dev/null") }'
```
gbooker, line 10: This may be better patterned after the init.d file's stop method which is:
```
# Ask nicely
pids="$(ps -ef | grep 'Plex Media Server' | grep -v grep | awk '{print $2}')"
kill -15 $pids

sleep 5

# Stuck
pids="$(ps -ef | grep /usr/lib/plexmediaserver | grep -v grep | awk '{print $2}')"

if [ "$pids" != "" ]; then
  kill -9 $pids
  sleep 2
fi
```

NateTheGreat68: That's a good suggestion; simpler scripting without the tail --pid mess. Since it's guaranteed to take between 5 and 7 seconds to exit, and finish scripts are by default killed by s6 after 5 seconds, I believe it would also need a timeout-finish file in the service directory with contents of just "8000" to give it 8 seconds. That's not a problem though.

NateTheGreat68: I made the suggested edits; they're reflected in commit https://github.com/plexinc/pms-docker/pull/53/commits/60d3efc28119f9ec30fbca6d3334a0ff47872486. I'm going to mark this particular conversation as resolved for now.

PR #54, opened 2020-05-27
Add notes for composing Docker for ARM

therealklanni:
Was running into issues using docker-compose and there's no issue tracker here, so I had to figure it out myself. Thought the solution could be useful for others. 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,9 +86,11 @@ Note: In this configuration, you must do some additional configuration:
 - If you wish your Plex Media Server to be accessible outside of your home network, you must manually setup port forwarding on your router to forward to the `ADVERTISE_IP` specified above.  By default you can forward port 32400, but if you choose to use a different external port, be sure you configure this in Plex Media Server's `Remote Access` settings.  With this type of docker networking, the Plex Media Server is essentially behind two routers and it cannot automatically setup port forwarding on its own.
 - (Plex Pass only) After the server has been set up, you should configure the `LAN Networks` preference to contain the network of your LAN.  This instructs the Plex Media Server to treat these IP addresses as part of your LAN when applying bandwidth controls.  The syntax is the same as the `ALLOWED_NETWORKS` below.  For example `192.168.1.0/24,172.16.0.0/16` will allow access to the entire `192.168.1.x` range and the `172.16.x.x` range.
 
-### ARM notes
+### Using `docker-compose` on ARM devices
 
-To use the docker-compose scripts with ARM (SBCs like Raspberry Pi), you must first build the docker image locally.
+The provided `docker-compose` templates use the `plexinc/pms-docker` image which is the amd64 build and won't work on ARM devices.
+
+To use `docker-compose` with ARM devices, you must first build one of the ARM images locally and make a small modification to your chosen template, as shown below.
 
 ```sh
 docker build -f Dockerfile.armv7 # or arm64
--- a/README.md
+++ b/README.md
@@ -90,27 +90,10 @@ Note: In this configuration, you must do some additional configuration:
 
 The provided `docker-compose` templates use the `plexinc/pms-docker` image which is the amd64 build and won't work on ARM devices.
 
-To use `docker-compose` with ARM devices, you must first build one of the ARM images locally and make a small modification to your chosen template, as shown below.
+To use `docker-compose` with ARM devices, you must first build one of the ARM images locally.
 
 ```sh
-docker build -f Dockerfile.armv7 # or arm64
-```
-
-Then, remove the `image: plexinc/pms-docker` line from your chosen compose template, and add the line `build: .` in its place.
-
-E.g.
-
-```diff
-@@ -2,7 +2,7 @@
- version: '2'
- services:
-   plex:
-     container_name: plex
--    image: plexinc/pms-docker
-+    build: .
-     restart: unless-stopped
-     ports:
-       - 32400:32400/tcp
+docker build -t plexinc/pms-docker:latest -f Dockerfile.armv7 # or arm64
 ```
 
 Then you can `docker-compose up`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,6 +86,33 @@ Note: In this configuration, you must do some additional configuration:
...
 
+### ARM notes
+
+To use the docker-compose scripts with ARM (SBCs like Raspberry Pi), you must first build the docker image locally.
+
+```sh
+docker build -f Dockerfile.armv7 # or arm64
+```
+
+Then, remove the `image: plexinc/pms-docker` line from your chosen compose template, and add the line `build: .` in its place.
```
gbooker, line 97: If you build with `-t plexinc/pms-docker:latest` then this (and the below) is completely unnecessary.

README.md
```diff
@@ -86,6 +86,33 @@ Note: In this configuration, you must do some additional configuration:
 - If you wish your Plex Media Server to be accessible outside of your home network, you must manually setup port forwarding on your router to forward to the `ADVERTISE_IP` specified above.  By default you can forward port 32400, but if you choose to use a different external port, be sure you configure this in Plex Media Server's `Remote Access` settings.  With this type of docker networking, the Plex Media Server is essentially behind two routers and it cannot automatically setup port forwarding on its own.
 - (Plex Pass only) After the server has been set up, you should configure the `LAN Networks` preference to contain the network of your LAN.  This instructs the Plex Media Server to treat these IP addresses as part of your LAN when applying bandwidth controls.  The syntax is the same as the `ALLOWED_NETWORKS` below.  For example `192.168.1.0/24,172.16.0.0/16` will allow access to the entire `192.168.1.x` range and the `172.16.x.x` range.
 
+### ARM notes
```
gbooker, line 89: If build instructions are provided, I'd prefer it to not be ARM specific but rather generic to all platforms.

README.md
therealklanni: I'm not sure what you mean. I was not able to get a working build with the standard image because it's not ARM compatible. Can you elaborate more?

README.md
therealklanni: The instructions are ARM-specific because there was not a working docker-compose template for ARM. It's just a couple extra details to help someone else know how to get a successful docker-compose working on ARM.

I considered making ARM-specific docker-compose templates, but it still required a note about running `docker build` locally anyway, so it made more sense (to me) to just add the template change to the notes as well, since it was such a small change.

README.md
therealklanni: For additional clarification, this is not about getting a "latest build" this is about building the **arm64/armv7** image (which are not provided though Docker Hub), just like you have to for ARM devices. The additional context here is how to do that *in conjunction with* using the provided `docker-compose` templates.

I can reword my changes to make that more obvious.

gbooker: I know exactly what you are trying to do; I'm telling you there's a better way:  When you run `docker build`, you can specify the tag under which it'll store the resulting image.  So, use the one expected by docker-compose in the existing template.  Then all you need is the build command.

therealklanni: Sorry, it seemed like you were saying not to build it yourself, I misread.

I'll update the instruction, thanks.

README.md
```diff
@@ -86,6 +86,18 @@ Note: In this configuration, you must do some additional configuration:
 - If you wish your Plex Media Server to be accessible outside of your home network, you must manually setup port forwarding on your router to forward to the `ADVERTISE_IP` specified above.  By default you can forward port 32400, but if you choose to use a different external port, be sure you configure this in Plex Media Server's `Remote Access` settings.  With this type of docker networking, the Plex Media Server is essentially behind two routers and it cannot automatically setup port forwarding on its own.
 - (Plex Pass only) After the server has been set up, you should configure the `LAN Networks` preference to contain the network of your LAN.  This instructs the Plex Media Server to treat these IP addresses as part of your LAN when applying bandwidth controls.  The syntax is the same as the `ALLOWED_NETWORKS` below.  For example `192.168.1.0/24,172.16.0.0/16` will allow access to the entire `192.168.1.x` range and the `172.16.x.x` range.
 
+### Using `docker-compose` on ARM devices
+
+The provided `docker-compose` templates use the `plexinc/pms-docker` image which is the amd64 build and won't work on ARM devices.
```
gbooker, line 91: Does it try to use the amd64 image?  I would have expected that it would just fail to fetch because there isn't an image for the arm architecture.

Assuming that this is correct, maybe it's better to say that the arm images aren't currently available on dockerhub so you have to build it manually.

therealklanni: When I tried it, yes, it built with the wrong arch.

gbooker: Huh, docker-compose built the (wrong architecture) image?  Was this when you previously had `build: .` in the file?  I'd be a bit surprised if it tries to build the image in its current form because that would mean it would have to find this repository in GH, clone it, and then build it.  Not entirely unreasonable for it to be able to do this but not what I would have expected.

therealklanni: No, I had taken the "bridge" docker-compose yaml over to my RasPi and tried to start it from that. The only edits I made were related to the env vars. It wouldn't start up, and when I checked I was getting a weird instruction error, which I soon realized was related to architecture. Didn't see any docs around it, but figured if I just built the armv7 myself and docker-compose using that it would be fine, and sure enough it was. That's why I created the PR.

therealklanni: To be clear I manually copied the Dockerfile-armv7 to the Pi to run the build, nothing was automatically cloned from GH. Not entirely sure what you meant there, sorry.

gbooker: That sounds like it grabbed an image of the wrong architecture.  That's very disappointing because the images do have labels indicating their architecture so docker has no excuse doing this.  Oh well.

therealklanni: Yeah, it was surprising, but I'm not a Docker expert, so I wasn't sure the reason behind it. I guess I kind of assumed Docker couldn't find the right architecture and defaulted to arm64?

PR #56, opened 2020-06-25
migrate to debian:stable-slim - smaller, newer.

dlgoodr:
discuss?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM debian:stable-slim
 
 ARG S6_OVERLAY_VERSION=v1.22.1.0
 ARG S6_OVERLAY_ARCH=amd64
@@ -17,7 +17,7 @@ RUN \
       curl \
       xmlstarlet \
       uuid-runtime \
-      unrar \
+      unrar-free \
     && \
 
 # Fetch and extract S6 overlay
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM debian:stable-slim
 
 ARG S6_OVERLAY_VERSION=v1.22.1.0
 ARG S6_OVERLAY_ARCH=aarch64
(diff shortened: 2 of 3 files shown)


gbooker:
I'd rather not get into an Ubuntu vs. Debian debate.

dlgoodr:
> I'd rather not get into an Ubuntu vs. Debian debate.

That's fair and I respect that - it isn't so much that I want to exit Ubuntu - all my hosts run Ubuntu! My Plex container runs on a VM running Ubuntu running on a KVM host running Ubuntu. `do-release-upgrade` is brilliant.

My concern is Ubuntu's Docker tagging strategy.  With Debian I can use a `stable-slim` tag and as new OSes come out I don't have to do anything.  Other OSes do this too but I selected Debian because it's the base of Ubuntu and should therefore be broadly compatible.

With Ubuntu, we're still on `16.04` here and it requires effort and attention to know all the things that need to be touched when 16.04 goes End of Life here Real Soon Now.  Ubuntu does not offer a `latest-lts` or similar tag so we cannot avoid this work if we stay with Ubuntu.

gbooker:
> My concern is Ubuntu's Docker tagging strategy. With Debian I can use a `stable-slim` tag and as new OSes come out I don't have to do anything.

This in intentionally not using such a tag.  There's testing that's required on a major OS upgrade.  Usually something doesn't break but we have to make sure.

> With Ubuntu, we're still on `16.04` here and it requires effort and attention to know all the things that need to be touched when 16.04 goes End of Life here Real Soon Now.

And I'll bump this to 20.04 before then (April 2021).  We are about a month away from 20.04.1 so that'd be an opportune time.

>  Ubuntu does not offer a `latest-lts` or similar tag so we cannot avoid this work if we stay with Ubuntu.

As stated above, this is intentionally not using such a tag.  Besides, they do offer such a tag.  From their docs:
> The `ubuntu:latest` tag points to the "latest LTS", since that's the version recommended for general use. The `ubuntu:rolling` tag points to the latest release (regardless of LTS status).

dlgoodr:
:woman_shrugging: fair.

PR #57, opened 2020-08-03
[S6 Overlay] Update the S6 Overlay to v2.0.1

damacus:
- Format the Dockerfiles
- Update the Overlay version to v2.0.0.1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:16.04
 
-ARG S6_OVERLAY_VERSION=v1.22.1.0
+ARG S6_OVERLAY_VERSION=v2.0.0.1
 ARG S6_OVERLAY_ARCH=amd64
 ARG PLEX_BUILD=linux-x86_64
 ARG PLEX_DISTRO=debian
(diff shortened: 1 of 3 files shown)


gbooker:
Any particular reason?
Also, why reformat the dockerfiles and thus destroy the intentional visual clues on command continuation?

damacus:
## S6

S6 made a few changes ([see here](https://github.com/just-containers/s6-overlay/compare/v1.22.1.0...v2.0.0.1) for the complete list), which are long term beneficial. However some of them are security related (tar balls having the wrong permissions), and some functionality like `/bin` linking which will be useful if/when the base image is upgrade past Ubuntu 16.04, as Ubuntu 19+ seems to have a problem with this. 

https://github.com/just-containers/s6-overlay#upgrade-notes

## Comments/Linting

This is probably partially personal preference, but I'm using the Docker extension for VSCode which uses the `[dockerfile-utils](https://github.com/rcjsuen/dockerfile-utils)` library to format

Given they're now inline, I'm actually finding it *easier* to read the entries.

gbooker:
As a consequence of updating containers to Ubuntu 20.04, their S6 versions were all updated to 2.1.  So as a result this PR is now defunct so I'm going to close.

PR #61, opened 2020-11-10
Only chown files that have incorrect user/group

yasn77:
Not all files will need to change user/group and doing for every file can take a
very long time for large libraries. This commit will use find to find files/dir
that have incorrect user/group and change them.

Hopefully, this should be quicker than going through and changing every single
file.

To be honest, I am still not sure about this approach as it still can take a while to complete.

The main issue for me is that I am using this image in Kubernetes and because there are liveness/readiness checks, it keeps getting recreated (readiness check is timing out). Ideally, I would like to see this maybe as a separate 'init' step, but I'll look at that and see if can create a new PR for it.

### DIFF
--- a/root/etc/cont-init.d/40-plex-first-run
+++ b/root/etc/cont-init.d/40-plex-first-run
@@ -67,10 +67,10 @@ fi
 if [ "${CHANGE_CONFIG_DIR_OWNERSHIP,,}" = "true" ]; then
   if [ -f "${prefFile}" ]; then
     if [ ! "$(stat -c %u "${prefFile}")" = "$(id -u plex)" ]; then
-      chown -R plex:plex /config
+      find /config \! \( -uid $(id -u plex) -gid $(id -g plex) \) -exec chown plex:plex {} \;
     fi
   else
-    chown -R plex:plex /config
+    find /config \! \( -uid $(id -u plex) -gid $(id -g plex) \) -exec chown plex:plex {} \;
   fi
   chown -R plex:plex /transcode
 fi


gbooker:
> Ideally, I would like to see this maybe as a separate 'init' step,

Docker doesn't really have a first-run type of setup (a serious deficiency if you ask me) and so this script does its best to emulate that via creation of a `/.firstRunComplete` file.  See the first few lines and last few lines of the script you modified.

yasn77:
@gbooker Yeh, I was thinking of using an environment variable (easier to inject into a container) that would allow skipping bits (or entire first-run).

Also, just a thought, but the location of `/.firstRunComplete` should probably be in `/config`

gbooker:
> Yeh, I was thinking of using an environment variable (easier to inject into a container) that would allow skipping bits (or entire first-run).

Uh, almost every aspect of it is already controlled by envs.

> Also, just a thought, but the location of `/.firstRunComplete` should probably be in `/config`

No, it's per creation of the container.  Using an existing `/config` for a new container should rerun aspects for the first-run.

yasn77:
> Uh, almost every aspect of it is already controlled by envs.

I see that, but my point was that currently there is no way to skip the `40-plex-first-run` script entirely unless there is a file at the root of the container.

So if you wanted to deploy plex in stages (useful for Kubernetes environment), using a single environment variable to skip running the script altogether would be helpful.

TBH this is a slight digression, the main concern I have at the moment is that the `chown` step can take a really long time with large libraries and my PR is an attempt to speed that up.

root/etc/cont-init.d/40-plex-first-run
```diff
@@ -67,10 +67,10 @@ fi
 if [ "${CHANGE_CONFIG_DIR_OWNERSHIP,,}" = "true" ]; then
   if [ -f "${prefFile}" ]; then
     if [ ! "$(stat -c %u "${prefFile}")" = "$(id -u plex)" ]; then
-      chown -R plex:plex /config
+      find /config \! \( -uid $(id -u plex) -gid $(id -g plex) \) -exec chown plex:plex {} \;
```
gbooker, line 70: Likely better to use `-print0` and pipe it into `xargs -0` as that's more efficient than `-exec`.

root/etc/cont-init.d/40-plex-first-run
```diff
@@ -67,10 +67,10 @@ fi
...
     fi
   else
-    chown -R plex:plex /config
+    find /config \! \( -uid $(id -u plex) -gid $(id -g plex) \) -exec chown plex:plex {} \;
```
gbooker, line 73: Same.

gbooker (commented):
Sorry I didn't get back to this earlier; got distracted.

I think this would be even better using xargs.  I checked and it is in the container so it should be safe to use.

gbooker:
Changes integrated into 716ed27

PR #62, opened 2020-12-08
Add Intel NEO driver

rcombs:


Dockerfile
```diff
@@ -27,6 +30,14 @@ RUN \
...
     tar xzf /tmp/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz -C /usr ./bin && \
     \
+# Fetch and install Intel Compute Runtime and its deps
+    curl -J -L -o '/tmp/gmmlib.deb' 'https://github.com/intel/compute-runtime/releases/download/${INTEL_NEO_VERSION}/intel-gmmlib_${INTEL_GMMLIB_VERSION}_amd64.deb' && \
+    apt install -y /tmp/gmmlib.deb && \
+    curl -J -L -o '/tmp/#1.deb' "https://github.com/intel/intel-graphics-compiler/releases/download/igc-${INTEL_IGC_VERSION}/{intel-igc-core,intel-igc-opencl}_${INTEL_IGC_VERSION}_amd64.deb" && \
+    apt install -y /tmp/intel-igc-core.deb /tmp/intel-igc-opencl.deb && \
+    curl -J -L -o /tmp/intel-opencl.deb https://github.com/intel/compute-runtime/releases/download/${INTEL_NEO_VERSION}/intel-opencl_${INTEL_NEO_VERSION}_amd64.deb && \
+    apt install -y /tmp/intel-opencl.deb && \
+    \
```
gbooker, line 40: Need to remove the debs from /tmp.

gbooker (changes requested):

Dockerfile
rcombs: Doesn't this already handle that? https://github.com/plexinc/pms-docker/blob/59ec1a92e60d5ef298aeb5c3a84ac3d3b158a796/Dockerfile#L56

gbooker: Ah yes; that's the same layer so that should work.

saltydk:
Will this ever be pushed to the public/plexpass/beta tags? Seems like an oversight.

gbooker:
New images have been pushed.

saltydk:
Thank you.

PR #63, opened 2020-12-10
add localhost mapping to `/etc/hosts`

Sawrz:
The health check for my container continually failed with the following message:
``` bash
curl: (6) Could not resolve host: localhost
```
I checked my `/etc/hosts` file and there was no `localhost` mapping. After running 
``` bash
echo "127.0.0.1 localhost" >> /etc/hosts
```
the health check was successful. 

I'm not sure if it's better to change `healthcheck.sh` or the `Dockerfile`, but I choose the `Dockerfile` since there may be other processes using `localhost`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,8 @@ RUN \
       beignet-opencl-icd \
       ocl-icd-libopencl1 \
     && \
+    echo "127.0.0.1 localhost" >> /etc/hosts
+    && \
     \
 # Fetch and extract S6 overlay
     curl -J -L -o /tmp/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz https://github.com/just-containers/s6-overlay/releases/download/${S6_OVERLAY_VERSION}/s6-overlay-${S6_OVERLAY_ARCH}.tar.gz && \


gbooker:
I'm a bit hesitant to make this change because of the different networking modes in which this container can be run.  I'm not sure this would always be correct if the user were to use host networking and changed the IP address of localhost to not be 127.0.0.1 (strangely some people do this).

I'm curious as to the exact circumstances which lead to your container not being able to resolve localhost already.  This is the first I've heard of this and the healthcheck has been there for quite some time so the problem doesn't appear to be common.

Sawrz:
I understand your point. I was surprised too to get this result during my health check. I didn't have any issues so far, even with older versions. It started to happen when I changed from Ubuntu Server to Arch for reasons.
So I got to think a bit about this issue. When using the network mode `host`, does it use the hosts' network configuration for initialization? To be more exact:
Does the container use `/etc/hosts` from the host machine? I have only a basic understanding of how Docker containers work, so I thought it's more or less encapsulated and uses its network configurations. But perhaps I'm wrong.

I checked my hosts `/etc/hosts` file, and `localhost` is not defined either. So maybe the issue is connected to my OS's configuration? It would make sense since the network mode's name is `host`...

gbooker:
> Does the container use `/etc/hosts` from the host machine?

So I tested this in a throwaway container (`docker run -it --network=host --rm ubuntu:20.04 /bin/sh`) and the `/etc/hosts` was a copy of my host's file (including modifications I made to the file).  When I did not use host networking, the `/etc/hosts` was clearly the Ubuntu default and contains an entry for `localhost`.

Given that, it's starting to sound like the real problem is that your host doesn't have a hosts entry for `localhost` and not a problem in the container.

Sawrz:
Yeah, the culprit was my hosts `/etc/hosts` file. Maybe it's worth to point that out in the documentation?

gbooker:
Added: 565a4de.

I'm going to close this as the better solution appears to be the user ensure the entry is in `/etc/hosts` for this and other reasons (a lot of things break if `localhost` can't be resolved).  If there is further reason for this PR, feel free to re-open.

PR #64, opened 2020-12-14
Fix bridge docker compose template

robbixc:
Hey guys, two little fixes regarding the `ALLOWED_NETWORKS` env variable usage

- template is incorrect because plex won't let you use customConnections
(via ADVERTISE_IP) if the given ip is not in the range of the
ALLOWED_NETWORKS
- "(plex pass only)" was incorrect, the ALLOWED_NETWORKS must be set
regardless, if you want to run the container in bridge mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ plexinc/pms-docker
 Note: In this configuration, you must do some additional configuration:
 
 - If you wish your Plex Media Server to be accessible outside of your home network, you must manually setup port forwarding on your router to forward to the `ADVERTISE_IP` specified above.  By default you can forward port 32400, but if you choose to use a different external port, be sure you configure this in Plex Media Server's `Remote Access` settings.  With this type of docker networking, the Plex Media Server is essentially behind two routers and it cannot automatically setup port forwarding on its own.
-- (Plex Pass only) After the server has been set up, you should configure the `LAN Networks` preference to contain the network of your LAN.  This instructs the Plex Media Server to treat these IP addresses as part of your LAN when applying bandwidth controls.  The syntax is the same as the `ALLOWED_NETWORKS` below.  For example `192.168.1.0/24,172.16.0.0/16` will allow access to the entire `192.168.1.x` range and the `172.16.x.x` range.
+- After the server has been set up, you should configure the `LAN Networks` preference to contain the network of your LAN.  This instructs the Plex Media Server to treat these IP addresses as part of your LAN when applying bandwidth controls.  The syntax is the same as the `ALLOWED_NETWORKS` below.  For example `192.168.1.0/24,172.16.0.0/16` will allow access to the entire `192.168.1.x` range and the `172.16.x.x` range.
 
 ### Using `docker-compose` on ARM devices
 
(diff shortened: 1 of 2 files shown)


docker-compose-bridge.yml.template
```diff
@@ -18,6 +18,7 @@ services:
       - TZ=<timezone>
       - PLEX_CLAIM=<claimToken>
       - ADVERTISE_IP=http://<hostIPAddress>:32400/
+      - ALLOWED_NETWORKS=<hostIPAddressCIDRRange>
```
gbooker, line 21: This has no connection with `customConnections` nor does it affect the ability to use it.  This is the pref `allowedNetworks` which has the description `List of IP addresses and networks that are allowed without auth`.

This is no way we are going to ever set a default value for `allowedNetworks`, EVER!  It introduces a security risk and even if it's just to the host, that's still unacceptable.

README.md
```diff
@@ -86,7 +86,7 @@ plexinc/pms-docker
...
-- (Plex Pass only) After the server has been set up, you should configure the `LAN Networks` preference to contain the network of your LAN.  This instructs the Plex Media Server to treat these IP addresses as part of your LAN when applying bandwidth controls.  The syntax is the same as the `ALLOWED_NETWORKS` below.  For example `192.168.1.0/24,172.16.0.0/16` will allow access to the entire `192.168.1.x` range and the `172.16.x.x` range.
+- After the server has been set up, you should configure the `LAN Networks` preference to contain the network of your LAN.  This instructs the Plex Media Server to treat these IP addresses as part of your LAN when applying bandwidth controls.  The syntax is the same as the `ALLOWED_NETWORKS` below.  For example `192.168.1.0/24,172.16.0.0/16` will allow access to the entire `192.168.1.x` range and the `172.16.x.x` range.
```
gbooker, line 89: This is taking about the pref `LAN Networks` which is only usable by those with a Plex Pass because it affect bandwidth controls (as the above text explicitly states).

Additionally the commit message talking about `ALLOWED_NETWORKS` is completely wrong.

gbooker:
Both commits are factually wrong so I'm going to close this on that basis.

docker-compose-bridge.yml.template
robbixc: Hi, thanks for taking a look at the PR :)
My suggestion to add the `ALLOWED_NETWORKS` var in the docker-compose template comes from the fact that without allowing the local network it is not possible for the client to connect to the server without auth.
Maybe a comment can be added to the file, recommending to set the variable. I believe it can throw off many people using plex without auth otherwise

gbooker: > recommending to set the variable

We will not recommend setting this variable because of the security vulnerability.  Furthermore we will not recommend using Plex without authentication by extension of the same.  We have seen too many instances where people have exposed their server without realizing that they had done it via this variable (and another that we since obliterated).

rcombs: > without allowing the local network it is not possible for the client to connect to the server without auth.

The recommended solution to this problem is, of course, to authenticate your clients. The allowed-networks pref is a legacy feature and its use isn't recommended to new users.

robbixc: I understand it can be a source of trouble if misused, I just believe the user should be presented with the full picture.

Use this template as it is -> plex authentication required.
OR
add the variable at your own risk, according to your network setup -> authentication not required to use plex.

Better to warn the user yourself instead of letting them dig in the preferences and figure out a solution without being aware of the consequences :sweat_smile:

PR #65, opened 2020-12-16
Add --force-architecture to dpkg installation command

marklr:
Mitigates an ARM installation issue:

```dpkg: error processing archive /tmp/plexmediaserver.deb (--install):
 package architecture (arm64) does not match system (armhf)
```

### DIFF
--- a/root/plex-common.sh
+++ b/root/plex-common.sh
@@ -71,6 +71,6 @@ function installFromRawUrl {
     exit 1
   fi
 
-  dpkg -i --force-confold /tmp/plexmediaserver.deb
+  dpkg -i --force-confold --force-architecture  /tmp/plexmediaserver.deb
   rm -f /tmp/plexmediaserver.deb
 }


gbooker:
Uh, doesn't that error mean you have the wrong package?  I don't expect installing `amd64` code on `armhf` is going to work very well.

marklr:
That's not amd64 but arm64...

gbooker:
Ah, yes, it is ARM, but still this does seem problematic.  What if you are not actually running a 64-bit processor then the ARM64 should still present problems trying to run.

So are you actually running the 64-bit ARM image (one that you built that is) and it is reporting inside the container as `armhf`?

marklr:
Yep, that's right. In some cases it's "aarch64", elsewhere it's "armhf".
Ultimately, building the image locally works fine but dpkg still thinks
there's a difference and refuses to proceed.

PR #66, opened 2021-02-10
Add --max-time 100 to healthcheck

putty182:
If Plex is running and accepting TCP connections on 32400 but is unable to send a response, curl will wait indefinitley until it does.

Adding `--max-time 60` sets an upper limit for how long curl will wait for a response.  If this is reached, curl will exit with an error status which lines up with the existing use of `--fail` to catch HTTP error responses.

### DIFF
--- a/root/healthcheck.sh
+++ b/root/healthcheck.sh
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 
 TARGET=localhost
-CURL_OPTS="--connect-timeout 15 --max-time 30 --silent --show-error --fail"
+CURL_OPTS="--connect-timeout 15 --max-time 100 --silent --show-error --fail"
 
 curl ${CURL_OPTS} "http://${TARGET}:32400/identity" >/dev/null
 
--- a/root/healthcheck.sh
+++ b/root/healthcheck.sh
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 
 TARGET=localhost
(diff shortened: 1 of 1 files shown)


gbooker:
Would be worth addressing the questions brought up in #29.

putty182:
HEALTHCHECK invokes a check every 5s, giving each invocation 2s to return successfully, and after 100s of failed checks will mark the container as unhealthy.  Note that HEALTHCHECK only invokes things; the 2s timeout won't actually terminate the invoked command:

```Dockerfile
HEALTHCHECK --interval=5s --timeout=2s --retries=20 CMD /healthcheck.sh || exit 1
```

The inclusion of `--connect-timeout 15` ensures that if Plex isn't accepting connections at all, the curl command will exit with an error code after 15s.  In such a scenario, the number of concurrent checks is effectively capped at 3:

```sh
#!/bin/sh -e

TARGET=localhost
CURL_OPTS="--connect-timeout 15 --silent --show-error --fail"

curl ${CURL_OPTS} "http://${TARGET}:32400/identity" >/dev/null
```

But if Plex *is* accepting connections and can get past the --connect-timeout gate, then the curl command has no upper bounds; each check spawned every 5s will run indefinately (until either Plex recovers and handles the queued requests, or someone/something kills the unhealthy container processes).

Including both `--connect-timeout 15` and `--max-time 100` effectively caps the total number of concurrent checks at 20 without impacting existing HEALTHCHECK behavour.

PR #67, opened 2021-06-12
upgrade ARM s6 to 2.2.0.0

stevenkaras:
The version bundled was 2.1.0.0, which fails to run on raspberry pi devices with the error:

```
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
s6-svscan: warning: unable to iopause: Invalid argument
s6-svscan: warning: executing into .s6-svscan/crash
[s6-init] ensuring user provided files have correct perms...s6-svscan panicked! Dropping to a root shell.
/bin/sh: 0: can't access tty; job control turned off
```

This was documented as a potential source of incompatibility in [nextcloud][0] and resolved the issue for me

[0]: https://github.com/crazy-max/docker-nextcloud/issues/61


gbooker (changes requested):
I don't see the connection that indicates that updating the S6 resolves the issue.

That aside, why is this PR only updating S6 on one platform and not all of them?  Also, why use 2.2.0.0 when 2.2.0.3 is the latest release version?

stevenkaras:
I found the linked issues by searching for the error message. Unfortunately, the error message doesn't give much to go off of aside from the problem being in s6 (or at least showing up there). The fix there was to upgrade s6, so I blindly applied that here.

There is no good reason to use 2.2.0.0 instead of 2.2.0.3, or to limit the change to armv7.

As such, I've bumped the version to 2.2.0.3 and applied the change to both arm platforms.

gbooker:
There's actually a third platform, AMD64, which should contain the same version of S6 unless there's a good reason for it not to.

So I added it there and just committed it all to `master` rather than go through another round here.

PR #68, opened 2021-10-12
Maybe use of Debian image would be better because lighter ?

tigerblue77:
Just an idea...
Don't hesitate to tell me if I'm wrong and why. Also, I removed "unrar" package as it is not available on default Debian repositories ; if it was necessary (I didn't see where) but "tar" command could be use without needing any package installation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM debian:bullseye
 
 ARG S6_OVERLAY_VERSION=v2.2.0.3
 ARG S6_OVERLAY_ARCH=amd64
@@ -20,7 +20,6 @@ RUN \
       curl \
       xmlstarlet \
       uuid-runtime \
-      unrar \
       beignet-opencl-icd \
       ocl-icd-libopencl1 \
     && \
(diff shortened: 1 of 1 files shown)


zbyte-king (approved):

Re4zOon:
Hi,

Just a quick question: Why are you using apt-get instead of apt?
Is it just preference, or?
I dont see the reason to use the old/outdated apt-get anymore, specially on fresh distros.

tigerblue77:
> Hi,
> 
> Just a quick question: Why are you using apt-get instead of apt? Is it just preference, or? I dont see the reason to use the old/outdated apt-get anymore, specially on fresh distros.

Hello,
"old/outdated `apt-get`" ? Sorry but I think you're wrong. `apt` is using `apt-get` and is just the recommended command for interactive use by administrators, not for use in shell scripts. That's why, when you use `apt` in a bash script you get the following message :
> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Re4zOon:
Sorry, you are right. I never really looked into the differences and whatnot.

Randyblo7 (changes requested):
Nice

PR #71, opened 2021-12-17
Add i386 support

Kisty:
First off, thanks for Plex and the Media Server and bringing it to Docker. I have managed to get Plex running on my old netbook so I set off to make a `Dockerfile` for i386. Would this be of interest?

### DIFF
--- a/Dockerfile.i386
+++ b/Dockerfile.i386
@@ -1,0 +1,60 @@
+FROM i386/ubuntu:18.04
+
+ARG S6_OVERLAY_VERSION=v2.2.0.3
+ARG S6_OVERLAY_ARCH=x86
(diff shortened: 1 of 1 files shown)


Dockerfile.i386
```diff
@@ -0,0 +1,60 @@
...
+
+RUN \
+# Update and get dependencies
+    apt-get update && \
+    apt-get install -y \
+      tzdata \
+      curl \
+      xmlstarlet \
+      uuid-runtime \
+      unrar \
```
gbooker, line 20: I see the OpenCL deps are missing.  I presume that it's not available on i386?

Kisty: Correct. I don't think `armv7` has it either

PR #72, opened 2022-01-12
Update README.md - add note about WSL2

kpolicano:
Is it safe to keep the `/config` directory on a Windows host if Docker is running on WSL2?

I've been posting to various Plex forums around the internet trying to get an answer to this question.  Nobody seems to know, so I'm asking here.  

Anecdotally, I have been running with `/config` on the WSL2 Windows host with for over a month with no corruption issues.  On Hyper-V, the database would become corrupt within a matter of hours or days.


gbooker:
It appears that WSL2 uses 9p for accessing the host's filesystem (FreeNAS looked into this a while back but found it lacking ACL features).  9p is capable of doing what is necessary for a database backend but that doesn't guarantee that MS actually implemented all of the pieces (such as all the file locking).  Without guarantees by MS that this is implemented, I will continue to highly discourage any use of docker where the config dir is stored on a Windows host even with WSL2.

kpolicano:
Thanks for the response.  Will close PR.

PR #74, opened 2022-04-27
add connectivity check to plex.tv on startup with wait and backoff

apnar:
At times on a cold boot there is race condition with my plex container starting and the network/dns being fully up.  When this happens the plex container gets stuck and never recovers once the network is fully up.

This PR adds a very simple check that the container can reach https://plex.tv before continuing its startup.  If it can't be reached it'll sleep 10 seconds and try again doubling the length of the sleep each time.

### DIFF
--- a/root/etc/cont-init.d/30-check-connectivity
+++ b/root/etc/cont-init.d/30-check-connectivity
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# verify conenctivity and sleep until established with backoff
+SLEEP_SECS=10
+echo "Verifying connectivity to https://plex.tv."
+until curl https://plex.tv -s -o /dev/null; do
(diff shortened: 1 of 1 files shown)


gbooker:
This strikes me as the wrong solution to the problem.  Instead of pausing the startup of the container indefinitely until the network is up, the better solution would be to keep it from getting "stuck" in startup.  Additionally, for anyone who's startup doesn't get "stuck" in startup without network present, putting this in the container would effectively render the container unusable until their internet connection is up.

Where exactly is it getting "stuck" and not recovering?

apnar:
Of course can't reproduce it now after it happened numerous time previously.  I agree that fixing the root issue would be better but this was a reasonably quick approach that was narrowly scoped in what it touched.  Anyway, I'll close this PR for now and reopen if I can get more specific details.

PR #75, opened 2022-06-23
Change base image from 20.04 to 22.04

acranox:


gbooker:
Is there a reason why this is desired right now?  Usually we've not updated this container's base until the same upgrade is offered for LTS users.  Canonical typically doesn't have users upgrade from one LTS to the next until version .1 of the new LTS is released.  Currently 22.04.1 is scheduled for August (https://wiki.ubuntu.com/Releases)

acranox:
Fair enough.  I didn't have a particular reason.

PR #77, opened 2023-01-07
Remove unused port 3005 from docker-compose template and readme

krsmanovic:
We want docker-compose and documentation aligned with Dockerfile commit https://github.com/plexinc/pms-docker/commit/61178f762e30348de2753cdad40df6e5b3ac6555

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,6 @@ docker run \
 -d \
 --name plex \
 -p 32400:32400/tcp \
--p 3005:3005/tcp \
 -p 8324:8324/tcp \
 -p 32469:32469/tcp \
 -p 1900:1900/udp \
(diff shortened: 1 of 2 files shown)


gbooker (changes requested):
There are 3 templates and this PR should modify all of them.

krsmanovic:
@gbooker I don't see port 3005 referenced in host and macvlan network driver templates. Maybe I am missing something?
https://github.com/plexinc/pms-docker/blob/766870b75cf72916e0930856d5a15e7f579bbe88/docker-compose-host.yml.template
https://github.com/plexinc/pms-docker/blob/766870b75cf72916e0930856d5a15e7f579bbe88/docker-compose-macvlan.yml.template

gbooker (approved):
Sorry, I was forgetting that those templates don't specify port numbers because they don't need them.

PR #80, opened 2023-02-25
feat: Move /.firstRunComplete to /tmp to prepare readonly rootfs

mkilchhofer:
I currently migrate all my Kubernetes workloads to containers with read-only rootfs.

The plex container is a bit tricky with the `/.firstRunComplete` file. If you'd move that file to `/tmp`, it would be possible to mount a temporary filesystem which is writable (Kubernetes `emptyDir`) below /tmp.

The s6 suite also have issues with read-only rootfs but it (officially) supports it by setting `S6_READ_ONLY_ROOT=1`. Ref: https://github.com/just-containers/s6-overlay#read-only-root-filesystem

### DIFF
--- a/root/etc/cont-init.d/40-plex-first-run
+++ b/root/etc/cont-init.d/40-plex-first-run
@@ -6,7 +6,7 @@ if [ "${DEBUG,,}" = "true" ]; then
 fi
 
 # If the first run completed successfully, we are done
-if [ -e /.firstRunComplete ]; then
+if [ -e /tmp/.firstRunComplete ]; then
   exit 0
 fi
 
@@ -136,5 +136,5 @@ if [ -z "$(getPref "TranscoderTempDirectory")" ]; then
   setPref "TranscoderTempDirectory" "/transcode"
 fi
 
-touch /.firstRunComplete
+touch /tmp/.firstRunComplete
 echo "Plex Media Server first run setup complete"


gbooker:
The `/.firstRunComplete` is used to indicate whether this is the first time the container is run or whether it is a subsequent run.  Moving this to `/tmp` will likely completely break this (because `/tmp` is by definition transitory and content there can be deleted without warning).

mkilchhofer:
I would be fine to move it completely elsewhere but having it inside the / (the root) is a really bad practice for the reason I mentioned above.
Are you fine if we move it elsewhere?

mkilchhofer:
@gbooker any update?

gbooker:
As stated, moving this indicator to `/tmp` will break functionality.  This PR hasn't changed from that.  Any change to this behavior would have to also not break the existing functionality.

Perhaps there's an id or something which doesn't change across updates which could be used to indicate that it is in the same effective container and this be stored in a more permanent location in `/config`?

mkilchhofer:
What ID do you mean? If you recreate the container from a container image (eg. on an update) its not the same container.

In which cases should the firstRun be triggered?

gbooker:
I was afraid the container id changes across updates.

Ideally, the firstRun should be triggered when a fresh container is created but not when a container is upgraded.  Additionally not when a configuration is changed (such as a volume map, etc)  It's a tricky thing to accomplish since docker is really missing a mechanism to do this.

saltydk:
What container upgrade are you talking about here? The root filesystem does not persist when a container is recreated (or upgraded as you state).

mkilchhofer:
As I am still affected of this issue, can we talk about it again?

@gbooker is your idea that this script is only executed when plex is initialized for the first time or every time Plex is upgraded (and thus the container is replaced from a new container image)?

mkilchhofer:
> I don't like the idea of modifying the container root image that much (as it is antithetical to how docker is supposed to work) but I couldn't find any good mechanism to accomplish this.

Its not a big deal. The real issue is, that this file should not live in a directory which contains other files. We can move it to wherever you want. Would you be okay with another location?

cilindrox:
Might be missing some context here, but what about a newer directory that we declare with the `VOLUME` directive in the dockerfile? This allows users to override it with persistent volumes and signals there's some sort of statefulness expected to operators

mkilchhofer:
> Might be missing some context here, but what about a newer directory that we declare with the `VOLUME` directive in the dockerfile? This allows users to override it with persistent volumes and signals there's some sort of statefulness expected to operators

That would be great!

root/etc/cont-init.d/40-plex-first-run
```diff
@@ -6,7 +6,7 @@ if [ "${DEBUG,,}" = "true" ]; then
 fi
 
 # If the first run completed successfully, we are done
-if [ -e /.firstRunComplete ]; then
+if [ -e /tmp/.firstRunComplete ]; then
```
gbooker, line 9: If moved to another directory, this if should check for the file existing on `/` to account for those where it has already run in this container before this file was moved to another directory.

root/etc/cont-init.d/40-plex-first-run
```diff
@@ -136,5 +136,5 @@ if [ -z "$(getPref "TranscoderTempDirectory")" ]; then
   setPref "TranscoderTempDirectory" "/transcode"
 fi
 
-touch /.firstRunComplete
+touch /tmp/.firstRunComplete
```
gbooker, line 139: If moving to another dir, this likely would need to run a `mkdir` for that dir first in case it doesn't already exist

gbooker (changes requested):
Moving to another directory would be fine (just not `/tmp` because of the meaning of that name) and users who wish to map a volume to that dir are welcome to do so.  Just the if needs to check `/` as well for containers where this has already run before moving the `firstRunComplete` file.

root/etc/cont-init.d/40-plex-first-run
mkilchhofer: This I do not understand. Upgrading from one container image to another (version upgrade) will definitively loose those file in `/.firstRunComplete`. There is definitely no edge case where this file would survive a container upgrade.
- If one would define a volume to `/`, plex would not work as the content of `/` would not be visible anymore before the entrypoint command would be executed
- You cannot upgrade a container from one image version to another image version without loosing data not living in a mapped volume

gbooker: Sorry, when I wrote this I was thinking of something else.

PR #82, opened 2023-06-27
feat: add a helm chart to deploy PMS Kubernetes 

MarshallAsch:
closes: https://github.com/plexinc/operations/issues/2886

Adds a new method to deploy a PMS container to a Kubernetes cluster. 

This chart will deploy a `StatefulSet` with a single instance of of PMS.
Data volumes can be mounted by creating `PersistentVolume`s or by using the `rclone` sidecars to be able to easily mount drives from arbitrary locations that may not have Kubernetes drivers.

By default this chart will create a PMS service _internal_ to the cluster that is published on the standard PMS port.
However, this will now allow direct external connections to PMS. 
A builtin ingress controller has been provided that will allow PMS to be accessed over the standard HTTP(s) ports, this will require manually reconfiguring the PMS instance to [advertise using this url](https://support.plex.tv/articles/200430283-network/).

PR #83, opened 2023-07-06
feat: add a workflow to package and release the helm chart

MarshallAsch:
In #82 a helm chart was added but it is not currently installable directly from helm without being packaged properly. 
This adds a github workflow that will automatically package and publish the chart using github pages. This workflow was modified from the one used by [ingress-nginx](https://github.com/kubernetes/ingress-nginx/blob/main/.github/workflows/helm.yaml)

Steps that are still needed:
- [ ] @lotooo create and set a `GITHUB_TOKEN` secret for this git repository
- [x] Create a `gh-pages` branch to store the artifacts
- [ ] enable github pages using the `gh-pages` branch on this repository

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+charts/** @plexinc/team-operations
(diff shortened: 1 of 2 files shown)


cilindrox (approved):
LGTM 👍

PR #88, opened 2023-07-21
fix: missed using the service account from the stateful set

MarshallAsch:


lotooo (approved):
It looks good when I render locally

PR #92, opened 2023-10-13
build: add chart-testing workflow

cilindrox:
Add a `ct` job to lint and test charts using helm/chart-testing. By default, any PRs that modify `charts/*` will be linted and the chart installed against a kind cluster to check for correctness.

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -1,0 +1,49 @@
+name: Lint and Test Charts
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_call:
+  workflow_dispatch:
+  pull_request:
+    paths:
+    - 'charts/**'
+
+jobs:
+  lint-test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Set up Helm
+      uses: azure/setup-helm@v3
+      with:
+        version: v3.13.1
+
+    - uses: actions/setup-python@v4
+      with:
+        python-version: 3.x
+        check-latest: true
+
+    - name: Set up chart-testing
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - 'charts/**'
+
+jobs:
+  lint-test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
MarshallAsch, line 18: I think fetch depth 0 is the default

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,49 @@
+name: Lint and Test Charts
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_call:
+  workflow_dispatch:
```
MarshallAsch, line 6: I dont think we need these triggers

MarshallAsch (approved):

.github/workflows/lint-test.yaml
cilindrox: Added these in case we want to call it manually/remotely (test post-merge/sanity checks)

.github/workflows/lint-test.yaml
cilindrox: > Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags. Refer [here](https://docs.github.com/actions/using-workflows/events-that-trigger-workflows) to learn which commit $GITHUB_SHA points to for different events.

ref: https://github.com/actions/checkout/#checkout-v4

`fetch-depth: 0` is required by `ct` in order to compute the diff against the main branch and figure out changes/tag bumps etc.

PR #95, opened 2023-11-14
build: fix ct lint steps

cilindrox:
Helm argument is redundant during lint and fails the checks.


cilindrox:
Currently blocking #94

PR #100, opened 2024-01-27
Helm chart: Define port for NodePort

alphayax:
# Initial problem

When installing the chart on my microk8s cluster, I use NodePort to make Plex accessible from internet. But, the port is randomly defined, and can change at any time.

# This PR Solution

Add the ability to define the node port to use when service is NodePort. Default is 32400.

- When service = NodePort, the nodePort will automatically be 32400 by default.
- If user want to override and define is own node port, he can define the `.Value.service.nodePort` value to whatever he want

### DIFF
--- a/charts/plex-media-server/values.yaml
+++ b/charts/plex-media-server/values.yaml
@@ -144,6 +144,9 @@ service:
   type: ClusterIP
   port: 32400
 
+  # Port to use when type of service is "NodePort" (32400 by default)
+  #nodePort: 32400
+
   # optional extra annotations to add to the service resource
   annotations: {}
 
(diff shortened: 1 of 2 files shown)


charts/plex-media-server/values.yaml
```diff
@@ -144,6 +144,9 @@ service:
   type: ClusterIP
   port: 32400
 
+  # Port to use when type of service is "NodePort" (32400 by default)
+  #nodePort: 32400
```
cilindrox, line 148: nit: indent spacing/level 

```suggestion
  # nodePort: 32400
```

cilindrox (approved):
LGTM 👍

cilindrox:
Thanks @alphayax - could you bump the version to a minor (`0.2.0`) so we can release this?

MarshallAsch (approved):

PR #101, opened 2024-02-25
Helm chart: Add the ability to use a pre existing peristent volume claim

jpflouret:
# Problem
I have an existing persistent volume and persistent volume claim with my PMS database. I want to use this pvc with the PMS helm chart.

# Solution
This PR uses a common helm pattern where the user can provide the name of a pre existing volume claim and the chart will configure the PMS container to use that claim instead of creating a volumeClaimTemplate.

### DIFF
--- a/charts/plex-media-server/templates/statefulset.yaml
+++ b/charts/plex-media-server/templates/statefulset.yaml
@@ -30,6 +30,11 @@ spec:
       priorityClassName: {{ .Values.priorityClassName | quote }}
       {{- end }}
       volumes:
+      {{- if .Values.pms.configExistingClaim }}
+      - name: pms-config
+        persistentVolumeClaim:
+          claimName: {{ .Values.pms.configExistingClaim | quote }}
+      {{- end }}
       - name: pms-transcode
         emptyDir: {}
       {{- if and .Values.rclone.enabled .Values.rclone.configSecret }}
@@ -164,6 +169,7 @@ spec:
       {{- with .Values.extraContainers }}
 {{- toYaml . | nindent 6 }}
       {{- end }}
+  {{- if not .Values.pms.configExistingClaim }}
   volumeClaimTemplates:
   - metadata:
       name: pms-config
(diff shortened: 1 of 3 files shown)


charts/plex-media-server/values.yaml
```diff
@@ -32,6 +32,10 @@ pms:
   # the volume size to provision for the PMS database
   configStorage: 2Gi
 
+  # Name of an existing `PersistentVolumeClaim` for the PMS database
+  # NOTE: When it's set configStorage and storageClassName are ignored
```
cilindrox, line 36: nit: wording.

```suggestion
  # NOTE: When set, 'configStorage' and 'storageClassName' are ignored.
```

charts/plex-media-server/templates/statefulset.yaml
```diff
@@ -30,6 +30,11 @@ spec:
       priorityClassName: {{ .Values.priorityClassName | quote }}
       {{- end }}
       volumes:
+      {{- if .Values.pms.configExistingClaim }}
+      - name: pms-config
+        persistentVolumeClaim:
+          claimName: {{ tpl .Values.pms.configExistingClaim . }}
```
cilindrox, line 36: Do we need to template the string here, or can we use the value directly instead?

```suggestion
          claimName: {{ .Values.pms.configExistingClaim | quote }}
```

cilindrox (commented):
LGTM - thanks @jpflouret - Could you bump the Chart.version a `minor` so we can release this?

jpflouret:
Suggestions applied. I could squash the commit with the review suggestions if you prefer that.

cilindrox (approved):
LGTM 👍 thanks @jpflouret !

I'll squash and merge this for release

jpflouret:
I suppose I need to sign the commit as per the branch rules. I'll ammed the commits with signature and force push to my branch.

cilindrox:
@jpflouret appreciate it!

jpflouret:
@cilindrox done!

PR #102, opened 2024-04-03
feat: allow setting pod annotations

d3adb5:
Currently there is no way to annotate the pods in PMS' StatefulSet. This adds a chart value to do precisely that. My use case is that I want to add Plex to my Linkerd service mesh, which can be done through the `linkerd.io/inject` annotation.


MarshallAsch (approved):

cilindrox (approved):

d3adb5:
> Thanks @d3adb5 - do you mind bumping the chart version (https://github.com/plexinc/pms-docker/blob/master/charts/plex-media-server/Chart.yaml#L25) to `v0.3.0` so we can release this?

Absolutely. Done in efa0b91. Thanks for the review. :+1:

PR #105, opened 2024-06-03
Add runtimeClassName

thomaspetit:
This option might be interesting for those that are running a GPU within their kubernetes cluster and want to be able to specify the runtimeClassName manually.

```
---
apiVersion: v1
kind: Pod
metadata:
  name: nbody-gpu-benchmark
  namespace: default
spec:
  restartPolicy: OnFailure
  runtimeClassName: nvidia
  containers:
  - name: cuda-container
    image: nvcr.io/nvidia/k8s/cuda-sample:nbody
    args: ["nbody", "-gpu", "-benchmark"]
    resources:
      limits:
        nvidia.com/gpu: 1
    env:
    - name: NVIDIA_VISIBLE_DEVICES
      value: all
    - name: NVIDIA_DRIVER_CAPABILITIES
      value: all
```

### DIFF
--- a/charts/plex-media-server/values.yaml
+++ b/charts/plex-media-server/values.yaml
@@ -83,6 +83,9 @@ initContainer:
   #
   #   echo "Done."
 
+# specify your own runtime class name eg use gpu
+runtimeClassName: "" 
+
 # the settings specific to rclone
 rclone:
   # if the rclone sidecar should be created
--- a/charts/plex-media-server/values.yaml
+++ b/charts/plex-media-server/values.yaml
@@ -84,7 +84,7 @@ initContainer:
   #   echo "Done."
 
 # specify your own runtime class name eg use gpu
-runtimeClassName: "" 
+runtimeClassName: ""
 
 # the settings specific to rclone
 rclone:
--- a/charts/plex-media-server/Chart.yaml
+++ b/charts/plex-media-server/Chart.yaml
@@ -22,7 +22,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.3.0
+version: 0.4.0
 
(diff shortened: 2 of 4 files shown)


cilindrox (approved):
LGTM 👍 - thanks @thomaspetit 

Could you bump the chart's version to `0.4.0`?

thomaspetit:
Done

thomaspetit:
@cilindrox can you have a look at the latest commit? I'd love to re-align my homelab to the latest version with gpu acceleration 😄

cilindrox (approved):
LGTM 👍

charts/plex-media-server/values.yaml
```diff
@@ -83,6 +83,9 @@ initContainer:
   #
   #   echo "Done."
 
+# specify your own runtime class name eg use gpu
+runtimeClassName: ""
```
cilindrox, line 87:
```suggestion
runtimeClassName: ""
```

thomaspetit:
I'm validating some local stuff and just noticed that the securityContext is not configureable which basically means that the GPU can be detected but not used by Plex. Probably need to make it configureable too.

cilindrox:
> I'm validating some local stuff and just noticed that the securityContext is not configureable which basically means that the GPU can be detected but not used by Plex. Probably need to make it configureable too.

Good call - we went with something a bit [too lenient](https://github.com/plexinc/pms-docker/blob/master/charts/plex-media-server/templates/statefulset.yaml#L158-L162) but maybe that whole block can be made configurable and the default value just uses the current settings?

LMK and I can keep this PR open

thomaspetit:
The security context that you're referring to ise indeed required for `rclone` but I you don't plan to use the rclone or gpu acceleration you maybe do not want to make your plex container privileged from a security pov.

I'll check how I can nicely make it fit and ensure we have something non-breaking 👍

cilindrox:
whoops, bad container - even more straightforward then ;)

thomaspetit:
🥳

PR #106, opened 2024-07-15
chore!: remove environment variables that are no longer used

MarshallAsch:


saltydk:
Since you've closed issues I just wanted to add a comment about updating the self-updating image tags as they are now 3 years old when you merge this.

gbooker (approved):

PR #107, opened 2024-07-16
feat: merge the amd64 and arm Dockerfiles

MarshallAsch:
Make use of the `TARGETARCH` and `TARGETPLATFORM` build args to build the `amd64`, `arm64`, and `arm/v7`  images from the same Dockerfile. 

This will also split-up and re-order some of the build steps to improve the build cache.


cilindrox (approved):
LGTM - maybe we should consider adding Hadolint or similar to start checking these as part of the CI step?

gbooker (approved):

PR #115, opened 2024-11-08
doc: remove outdated docker tags

Fmstrat:
The current `beta` and `public` tags on Docker Hub are over 4 years old, and using a very dated base image of `ubuntu` that has had many security patches since then.

Until these are added to a regular build process, they should be removed from the documentation.


Fmstrat:
@gbooker Why was this closed?

PR #116, opened 2024-12-01
added externalTrafficPolicy to service

koiralasandesh:
When NodePort is used, plex container is unable to determine the user IP because k8s does NAT.
Because of this, all traffic seems to originate from the cluster itself.
This breaks a bunch of plex functionalities like determining bitrate for transcode and many other that rely on determining if the user is in LAN or WAN.
Setting `externalTrafficPolicy: Local` will allow Plex to determine the actual IP of user because this disables k8s NAT.
But this setting comes with caveats: the pods can only be accessed by the specific Node IP where the pod is running.

Read more here: 
https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip
https://access.redhat.com/solutions/7028639

This is my first time contributing, please let me know if there is specific branching requirements. I could not find any guidelines to contributing. Thanks!

### DIFF
--- a/charts/plex-media-server/Chart.yaml
+++ b/charts/plex-media-server/Chart.yaml
@@ -22,7 +22,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.5.0
+version: 0.6.0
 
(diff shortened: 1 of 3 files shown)


cilindrox (approved):

cilindrox:
Thanks @koiralasandesh !

PR #117, opened 2024-12-23
Added liveness and readiness probe to pms container

koiralasandesh:
Please let me know if you would like me to include any default probes.

### DIFF
--- a/charts/plex-media-server/Chart.yaml
+++ b/charts/plex-media-server/Chart.yaml
@@ -22,7 +22,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.4.1
+version: 0.5.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/plex-media-server/Chart.yaml
+++ b/charts/plex-media-server/Chart.yaml
@@ -22,7 +22,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.5.0
+version: 0.6.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/plex-media-server/Chart.yaml
+++ b/charts/plex-media-server/Chart.yaml
@@ -3,9 +3,9 @@ name: plex-media-server
 description: A Helm chart for deploying a PMS server to a kubernetes cluster
 
 keywords:
-- plex
-- pms
-- Personal Media Server
+  - plex
+  - pms
+  - Personal Media Server
 
 home: https://www.plex.tv
 icon: https://www.plex.tv/wp-content/themes/plex/assets/img/favicons/plex-152.png
@@ -22,13 +22,13 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.6.0
+version: 0.7.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "1.16.0"
+appVersion: "1.41.3"
 
 sources:
-- https://github.com/plexinc/pms-docker
+  - https://github.com/plexinc/pms-docker
--- a/charts/plex-media-server/templates/statefulset.yaml
+++ b/charts/plex-media-server/templates/statefulset.yaml
@@ -119,6 +119,10 @@ spec:
           value: compute,video,utility
         {{- end }}
         {{- with .Values.pms.resources }}
+        livenessProbe:
+          {{- toYaml .Values.pms.livenessProbe | nindent 10 }}
+        readinessProbe:
+          {{- toYaml .Values.pms.readinessProbe | nindent 10 }}
         resources:
           limits:
           {{- with .limits }}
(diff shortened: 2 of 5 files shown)


charts/plex-media-server/Chart.yaml
```diff
@@ -22,13 +22,13 @@ type: application
...
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "1.16.0"
+appVersion: "1.41.3"
 
 sources:
-- https://github.com/plexinc/pms-docker
+  - https://github.com/plexinc/pms-docker
```
cilindrox, line 34:
```suggestion
- https://github.com/plexinc/pms-docker
```

charts/plex-media-server/Chart.yaml
```diff
@@ -3,9 +3,9 @@ name: plex-media-server
 description: A Helm chart for deploying a PMS server to a kubernetes cluster
 
 keywords:
-- plex
-- pms
-- Personal Media Server
+  - plex
+  - pms
+  - Personal Media Server
```
cilindrox, line 8: nit: no indentation on array keys.

```suggestion
- plex
- pms
- Personal Media Server
```

cilindrox (commented):
Thanks @koiralasandesh - let's include some default probes as comments underneath their `values.yaml` definition - that way, users can uncomment and expect a working probe by default.

charts/plex-media-server/Chart.yaml
```diff
@@ -22,13 +22,13 @@ type: application
...
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.6.0
+version: 0.7.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "1.16.0"
+appVersion: "1.41.3"
```
cilindrox, line 31: @MarshallAsch does this app bump look good or should this match the docker tag (currently `4.141.0` if I'm not mistaken)

koiralasandesh: I changed it to `1.41.3` which appears to be the current `latest` pms docker image version. If I understand correctly,  `4.141.0` seems like plex-web version.

Either way, please let me know what is expected and I can change it.

koiralasandesh:
> Thanks @koiralasandesh - let's include some default probes as comments underneath their `values.yaml` definition - that way, users can uncomment and expect a working probe by default.

@cilindrox can you please provide me some guidance on what you think would be good defaults for these probes?

I am personally using this in my values file, would you like me to add this as default or update anything?

`
  livenessProbe:
    httpGet:
      path: /identity
      port: 32400
    initialDelaySeconds: 60
    periodSeconds: 60
    timeoutSeconds: 1
    failureThreshold: 3
  readinessProbe: null
`

charts/plex-media-server/Chart.yaml
cilindrox: Can confirm: https://hub.docker.com/layers/plexinc/pms-docker/1.41.3.9314-a0bfb8370/images/sha256-19746abbf989b6be7b5798149567ef4c498a6a848b47f53d0dd948078b32eb0a - thanks for clarifying.

cilindrox:
> > Thanks @koiralasandesh - let's include some default probes as comments underneath their `values.yaml` definition - that way, users can uncomment and expect a working probe by default.
> 
> @cilindrox can you please provide me some guidance on what you think would be good defaults for these probes?
> 
> I am personally using this in my values file, would you like me to add this as default or update anything?
> 
> ` livenessProbe: httpGet: path: /identity port: 32400 initialDelaySeconds: 60 periodSeconds: 60 timeoutSeconds: 1 failureThreshold: 3 readinessProbe: null`

sounds good - either `/identity` or `/web` work here - maybe we can do `/web` for readiness?

cilindrox (approved):

gbooker:
I'm not certain of the definitions of liveness and readiness but `/identity` and `/web` are not going to behave that differently.  Both are going to be fairly lightweight but `/web` will yield a redirect which may not be desirable.

Note: when PMS is starting up, running database migrations, and essentially not quite ready to respond to HTTP requests, these endpoint will all respond but with a 503.  While there are some other cases where a 503 is returned on endpoints, this is the only case it is returned on these two.  This would be better to make sure it doesn't result in a container being restarted while yielding a 503 because if it's a particularly long db migration, the restart will just result in it being run again from the beginning and will indefinitely until the migration is allowed enough time to complete.

cilindrox:
@gbooker the restart scenario would be covered by the readiness checks, then the control is handed to the liveness probes. Good point about the redirect too, guess `/identity` should be used for both in that case.

MarshallAsch (approved):
looks good to me, same comment as Gaston about the commented defaults. 

Also re-run the command [helm-docs](https://github.com/norwoodj/helm-docs) to update the readme with the new values, (probably after you add the comment)

koiralasandesh:
Hello @cilindrox I have added commented liveness and readiness checks as per our discussion.

However, I am not much familiar with helm-docs. I installed [helm-docs](https://github.com/norwoodj/helm-docs) as per @MarshallAsch's instructions and tried to generate the docs. But, the generated one was quite different from the one in the repo. It only has the values table and they are missing the comment descriptions. 

Can you point me to an instruction or Readme that might help me understand how I can update the docs?

MarshallAsch:
I'll take care of the docs then, I'll push a commit tomorrow probably. (I had a work in progress to switch the doc tool I might have mixed up if I had merged that one or not)

cilindrox (approved):
LGTM 👍 - thanks @koiralasandesh

charts/plex-media-server/templates/statefulset.yaml
```diff
@@ -119,6 +119,10 @@ spec:
           value: compute,video,utility
         {{- end }}
         {{- with .Values.pms.resources }}
+        livenessProbe:
```
jpflouret, line 122: It appears to me that this is incorrectly nested within resources due to the above line.

PR #118, opened 2024-12-31
Fix incorrect nesting of liveness and readiness probes

jpflouret:
Liveness and readiness probes introduced in #117 are incorrectly nested within the `resources` template. Adding resource limits to the pms chart results in the following error:

```
Error: template: plex-media-server/templates/statefulset.yaml:123:28: executing "plex-media-server/templates/statefulset.yaml" at <.Values.pms.livenessProbe>: nil pointer evaluating interface {}.pms
```

This change fixes that issue.

### DIFF
--- a/charts/plex-media-server/Chart.yaml
+++ b/charts/plex-media-server/Chart.yaml
@@ -22,7 +22,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.7.0
+version: 0.7.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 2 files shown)


MarshallAsch (approved):

cookandy (approved):

jpflouret:
I believe the checks are failing due to the fact that the liveness and readyness probes are by default empty which is not valid.

```
        livenessProbe:
          {}
        readinessProbe:
          {}
```

I think this error didn't materialize itself in the previous PR due to it being hidden by the `resources` error that this PR fixes.

MarshallAsch (approved):

jpflouret:
@MarshallAsch I think your commit is unsigned. Do you want me to redo the chart bump change and sign it myself?

MarshallAsch:
no thats okay. I havent fixed my signing key so my commits are all unsigned now (I can bypass that requirement)

PR #120, opened 2025-03-02
Add support for extra init containers and sidecar containers

RegentOfOrigin:
The current chart asserts that only a single init container is needed for rclone. This can be adapted for most purposes so long as the specified image contains an entrypoint at /init/init.sh and is meant to be run as an initContainer and not a sidecar container (https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#api-for-sidecar-containers).

There are useful tools that operate on plex's metadata files, bundles, or database that would be best run as sidecar containers (see above), which are functionally init containers with restart policies.

This pull request adds support for additional init containers through an "extraInitContainers" argument that is semantically identical to the existing extraContainers argument.

### DIFF
--- a/charts/plex-media-server/templates/statefulset.yaml
+++ b/charts/plex-media-server/templates/statefulset.yaml
@@ -96,6 +96,9 @@ spec:
         - name: rclone-config
           mountPath: /out
       {{- end }}
+      {{- with .Values.extraInitContainers }}
+{{- toYaml . | nindent 6 }}
+      {{- end }}
       containers:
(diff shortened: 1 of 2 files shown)


cilindrox (approved):
LGTM - thanks @RegentOfOrigin

cilindrox:
cc @MarshallAsch for a quick spot check

MarshallAsch (approved):
needs a `minor` version bump of the chart version and its good. (we can do that tho after merging)

RegentOfOrigin:
Awesome, thank you so much!

PR #123, opened 2025-03-08
[helm chart] Allow the pms container to run as privileged

Pale-whale:
Hello

I've try to make hardware transcoding with an intel igpu work as a non privileged container with no success.
If it is possible I'd like some pointers on how to do that
If not this pr adds the ability to run the pms container as privileged

Thanks for providing an official chart 🙏

### DIFF
--- a/charts/plex-media-server/templates/statefulset.yaml
+++ b/charts/plex-media-server/templates/statefulset.yaml
@@ -126,11 +126,13 @@ spec:
         readinessProbe:
           {{- toYaml . | nindent 10 }}
         {{- end }}
+        {{- if .Values.pms.privileged }}
         securityContext:
           privileged: true
           capabilities:
             add:
               - SYS_ADMIN
+        {{- end }}
         {{- with .Values.pms.resources }}
         resources:
           limits:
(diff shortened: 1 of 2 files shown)


charts/plex-media-server/templates/statefulset.yaml
```diff
@@ -126,6 +126,13 @@ spec:
         readinessProbe:
           {{- toYaml . | nindent 10 }}
         {{- end }}
+        {{- if .Values.pms.privileged }}
+        securityContext:
+          privileged: true
+          capabilities:
+            add:
+              - SYS_ADMIN
+        {{- end }}
```
cilindrox, line 135: Instead of the single capability being added here, I'd rather we did the whole securityContext stanza, eg:

```suggestion
        {{- with .Values.securityContext }}
        securityContext:
          {{- toYaml . | nindent 10 }}
        {{- end }}
```

And then:

```yaml
# values.yaml

securityContext: {}
```

cilindrox (commented):
Thanks for the PR @Pale-whale - left some comments - IMO we might take this opportunity to make the whole `securityContext` configurable.

o10s:
@cilindrox thank you for doing this I could finally get rid of Intel GPU device plugin which is a hog to setup and use !
could someone approve this PR pretty please !
thank you !

cilindrox:
This was implemented in #125

PR #124, opened 2025-03-21
feat: use fixed tag instead of latest in helm chart

tzabbi:
Since a helm chart is made for a specific version of the app, we should specify the image with that version instead of using latest


cilindrox (approved):
LGTM 👍 - thanks @tzabbi

PR #129, opened 2025-03-22
Update alpine and rclone image tags

tzabbi:
Update rclone container image version from 1.62 to 1.69 and alpine init container from 3.18 to 3 which is using the latest minor version of alpine. 

Alternatively we can use 3.21 but I would appreciate if you add dependabot or renovate.

### DIFF
--- a/charts/plex-media-server/values.yaml
+++ b/charts/plex-media-server/values.yaml
@@ -4,7 +4,7 @@ image:
   registry: index.docker.io
   repository: plexinc/pms-docker
   # -- If unset use "latest"
-  tag: "latest"
+  tag: "1.41.5.9522-a96edc606"
   sha: ""
   pullPolicy: IfNotPresent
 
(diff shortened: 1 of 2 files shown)


charts/plex-media-server/values.yaml
```diff
@@ -93,7 +93,7 @@ initContainer:
     registry: index.docker.io
     repository: alpine
     # -- If unset use latest
-    tag: 3.18.0
+    tag: 3
```
cilindrox, line 96: nit: I think I'd prefer pinning to the full semver and not just a moving tag

cilindrox (commented):
thanks @tzabbi - this LGTM - can you pin that alpine image using the latest patch version instead?

charts/plex-media-server/values.yaml
tzabbi: Done.

cilindrox:
thanks @tzabbi !

PR #137, opened 2025-04-28
feat: add service.loadBalancerIP

briannosaurus:
Added the loadBalancerIP key so people using metalLB can set the service ip address that way


cilindrox:
Thanks @briannosaurus - merging and releasing this as part of v0.10

PR #138, opened 2025-04-28
Update statefulset.yaml

echo-devnull:
Also see: https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html

And: https://forums.plex.tv/t/claiming-a-plex-media-server-docker-kubernetes/914912/24

Simply put: Kubernetes is trying to internally resolve "plex.tv" because it thinks this is a internal service name (only 1 dot).

We can either make plex media server try and resolve the actual FQDN for `plex.tv.` (ending with a dot)

Or add this dnsconfig to the statefulset.

### DIFF
--- a/charts/plex-media-server/templates/statefulset.yaml
+++ b/charts/plex-media-server/templates/statefulset.yaml
@@ -21,6 +21,10 @@ spec:
       annotations:
         {{- toYaml .Values.statefulSet.podAnnotations | nindent 8 }}
     spec:
+      dnsConfig:
+        options:
+          - name: ndots
+            value: "1"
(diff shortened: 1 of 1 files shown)


cilindrox:
Thanks @echo-devnull - I've added a configurable `dnsConfig` stanza for v0.10 (#141)
Users should be able to configure ndots from there.

cilindrox:
[v0.10.0](https://github.com/plexinc/pms-docker/releases/tag/helm-chart-0.10.0) is out now with a configurable `dnsConfig` stanza. Thanks @echo-devnull

PR #139, opened 2025-05-02
Add more flexible TLS configuration

iPenguin:
This is a breaking change to the configuration, but it allows for a more flexible TLS configuration.

The current configuration doesn't allow you to configure TLS and also leave the secretName option un-configured which allows you to use the ingress' default TLS certificate.

### DIFF
--- a/charts/plex-media-server/values.yaml
+++ b/charts/plex-media-server/values.yaml
@@ -24,11 +24,10 @@ ingress:
 
   # -- Optional TLS configuration to provide valid https connections
   # using an existing SSL certificate
-  # tls:
+  tls: []
   #   - hosts:
   #       - plex.example.com
-  #     secretName: ""
-  tls: []
+  #     secretName: cert-example-com
 
   # -- Custom annotations to put on the ingress resource
   annotations: {}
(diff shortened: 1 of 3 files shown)


charts/plex-media-server/values.yaml
```diff
@@ -22,9 +22,13 @@ ingress:
...
 
-  # -- Optional secret name to provide valid https connections
+  # -- Optional TLS configuration to provide valid https connections
   # using an existing SSL certificate
-  certificateSecret: ""
+  # tls:
+  #   - hosts:
+  #       - plex.example.com
+  #     secretName: ""
+  tls: []
```
cilindrox, line 31: nit: place example config values below the var definition.

```suggestion
  tls: []
  #   - hosts:
  #       - plex.example.com
  #     secretName: example-dot-com
```

cilindrox (commented):
LGTM - thanks @iPenguin! Could you bump the [Chart version](https://github.com/plexinc/pms-docker/blob/master/charts/plex-media-server/Chart.yaml#L25) to 1.0.0 ?

charts/plex-media-server/values.yaml
iPenguin: Hi @cilindrox thanks for the feedback I've pushed another commit with your suggestions.

-Brian

cilindrox (approved):

cilindrox:
thanks @iPenguin - v1.0 is out now!

PR #144, opened 2025-05-08
Fixing indenting of tls config in ingress template

RobReus:
With the v1 change of the helm chart, the indenting of the tls block shifted right by 2 spaces, effectively removing the tls config altogether. This fixes that by shifting the tls block to the left again by 2 spaces.


RobReus:
> Thanks @RobReus Do you mind bumping the `Chart.yaml`s version to v1.0.1?

Done :)

PR #148, opened 2025-05-23
Update Dockerfile to be a supported ubuntu version

clintonsteiner:


cilindrox:
cc @gbooker - you might have some more insight as to whether we can support this

gbooker:
Merged.  It didn't want to allow rebase/merge because of requiring signed commits so I did a squash/merge.

PR #154, opened 2025-07-28
fix: if statement breaks "env"

Xeroxxx:
The statement introduced breaks the helmchart. If ingress or ADVERTISE_IP is not defined the chart fails as "env:" is missing.


Xeroxxx:
@cilindrox Please review.

cilindrox (approved):
Good catch, thanks @Xeroxxx !

PR #169, opened 2026-05-18
fix: guard Plex finish kill when no PID is found

OmarB97:
## Summary

Guard the initial `kill -15 $pids` call in `root/etc/services.d/plex/finish` so it only runs when a Plex Media Server PID was found.

## Why

If Plex has already exited by the time the finish script runs, `pids` is empty and the shell executes `kill -15` without a PID. That prints the `kill: usage...` message during container shutdown. The later SIGKILL block already guards against an empty PID list, so this makes the TERM path match the existing stuck-process path.

## Validation

- `bash -n root/etc/services.d/plex/finish`
- Reproduced in a downstream Unraid container using `plexinc/pms-docker:latest`, where the shutdown logs showed `kill: usage...`; applying the guard removed that message while Plex restarted healthy.

### DIFF
--- a/root/etc/services.d/plex/finish
+++ b/root/etc/services.d/plex/finish
@@ -3,10 +3,22 @@
 
 echo "Stopping Plex Media Server."
 
+kill_pids() {
+  local signal="$1"
+  local pids="$2"
+  local pid
+
+  while IFS= read -r pid; do
+    if [ "$pid" != "" ]; then
+      kill "$signal" "$pid"
+    fi
+  done <<< "$pids"
+}
+
 # Ask nicely
 pids="$(ps -ef | grep 'Plex Media Server' | grep -v grep | awk '{print $2}')"
 if [ "$pids" != "" ]; then
-  kill -15 $pids
+  kill_pids -15 "$pids"
 fi
 
 sleep 5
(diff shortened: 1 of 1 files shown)


root/etc/services.d/plex/finish
```diff
@@ -5,7 +5,9 @@ echo "Stopping Plex Media Server."
 
 # Ask nicely
 pids="$(ps -ef | grep 'Plex Media Server' | grep -v grep | awk '{print $2}')"
-kill -15 $pids
+if [ "$pids" != "" ]; then
+  kill -15 $pids
```
cilindrox: Can we quote the `$pids` var for safety here and on L18 below?

```suggestion
  kill -15 "$pids"
```

cilindrox (commented):
Thanks @OmarB97 - some comments regarding improving the globbing/splitting of the unquoted vars.

OmarB97:
> Thanks @OmarB97 - some comments regarding improving the globbing/splitting of the unquoted vars.

Thanks, pushed c506e49 to address this.

I quoted the PID handling, but used a small `kill_pids` helper instead of calling `kill -15 "$pids"` directly because the command substitution can return multiple newline-separated PIDs. Quoting the whole value in a direct `kill` call would turn multiple PIDs into one argument; the helper reads each PID line and calls `kill "$signal" "$pid"`, so each PID is quoted safely while preserving multi-process behavior.

I applied the same path for both the SIGTERM and SIGKILL calls and checked it with `bash -n` plus a stubbed multi-PID helper test.

PR #171, opened 2026-07-27
feat: add PLEX_MANUAL_PORT env var for manual Remote Access port mapping

jonfen:
## Summary

- Adds a `PLEX_MANUAL_PORT` environment variable that sets `ManualPortMappingMode=1` and `ManualPortMappingPort=<value>` in `Preferences.xml`, equivalent to enabling **Remote Access > Manually specify public port** in the server settings.
- Implemented in `40-plex-first-run` directly alongside the existing `ADVERTISE_IP` / `ALLOWED_NETWORKS` handling, using the same `setPref` helper and the same first-run semantics.
- Documents the new variable in the README parameters list, and extends the Bridge-networking note (which already explains that automatic port mapping cannot work there) to point at it.

*(Note: this repository has issues disabled, so the problem report below is included here rather than as a linked issue.)*

## Problem

When Plex Media Server runs behind a router/firewall that does not offer UPnP/NAT-PMP (most business-grade firewalls, and effectively any Bridge-networking deployment, where the README notes the server is "essentially behind two routers"), automatic port mapping can never succeed. Unless the user manually enables **Settings > Remote Access > Manually specify public port** in the web UI, the server never publishes a direct WAN connection to plex.tv — even when a correct port-forward for 32400 is already in place and working.

The failure mode is silent and easy to misdiagnose: remote clients can complete their small candidate-test connections against the forwarded port, but the candidate list plex.tv serves them contains no direct WAN URI, so every remote stream falls back to the bandwidth-limited relay (~2 Mbps). Users see "speed issues" while their firewall shows the forwarded port open and passing traffic.

Confirmed on `plexinc/pms-docker:latest` (PMS 1.43.3): with a static DNAT for 32400 in place, `/myplex/account` showed `mappingState="mapped"` but `publicPort="0"`, and the plex.tv resources list had only LAN, custom, and relay candidates. Setting `ManualPortMappingMode=1` + `ManualPortMappingPort=32400` immediately produced the `https://<wan-ip-dashes>.<hash>.plex.direct:32400` candidate and remote clients switched from `relayed="1"` to `relayed="0"`.

Today the only fix is a manual, non-declarative GUI step that is lost if `Preferences.xml` is rebuilt. The container already maps `ADVERTISE_IP` and `ALLOWED_NETWORKS` onto `Preferences.xml`, and this image is *the* deployment shape where automatic mapping is most likely to be impossible, so this belongs alongside them.

## Test plan

- [x] Verified on a live deployment (PMS 1.43.3, bridge networking, firewall without UPnP, static 32400 DNAT): applying these two preferences flipped `/myplex/account` from `publicPort="0"` to `publicPort="32400"`, added the direct `plex.direct:32400` WAN candidate to the plex.tv resources list, and remote clients switched from `relayed="1"` to `relayed="0"`.
- [ ] Fresh-container first run with `-e PLEX_MANUAL_PORT=32400` writes both attributes to a clean `Preferences.xml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ plexinc/pms-docker
 
 Note: In this configuration, you must do some additional configuration:
 
-- If you wish your Plex Media Server to be accessible outside of your home network, you must manually setup port forwarding on your router to forward to the `ADVERTISE_IP` specified above.  By default you can forward port 32400, but if you choose to use a different external port, be sure you configure this in Plex Media Server's `Remote Access` settings.  With this type of docker networking, the Plex Media Server is essentially behind two routers and it cannot automatically setup port forwarding on its own.
+- If you wish your Plex Media Server to be accessible outside of your home network, you must manually setup port forwarding on your router to forward to the `ADVERTISE_IP` specified above.  By default you can forward port 32400, but if you choose to use a different external port, be sure you configure this in Plex Media Server's `Remote Access` settings or via the `PLEX_MANUAL_PORT` environment variable (see below).  With this type of docker networking, the Plex Media Server is essentially behind two routers and it cannot automatically setup port forwarding on its own, so setting `PLEX_MANUAL_PORT` is recommended to ensure remote clients connect directly instead of through the bandwidth-limited relay.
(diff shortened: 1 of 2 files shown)


cilindrox (commented):
thanks @jonfen - what about renaming `PLEX_MANUAL_PORT` to `PLEX_PORT` and defaulting to the well-known `32400`? 

The script could then check if the var has changed from the defaults and set `ManualPortMapping*`.

That way, the user gets to configure the port, and we retain the existing functionality by default. Thoughts?

PR #175, opened 2026-08-23
feat(helm): add podSecurityContext

matlaj:
`pms.securityContext` is documented as "Security context for PMS pods", but it renders into the container's securityContext, and the chart has no pod-level securityContext hook at all.

That mismatch silently discards pod-only fields. Setting `fsGroup` -- the documented way to make the config volume group-writable -- produces a manifest where `fsGroup` sits in a container securityContext, which the API server drops. The chart installs cleanly and the setting simply never takes effect.

Add `pms.podSecurityContext`, rendered on the pod spec, and correct both doc strings to say which level they apply to.

Backward compatible: `pms.securityContext` keeps rendering exactly where it did, and the new value defaults to `{}`, which emits nothing.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/charts/plex-media-server/Chart.yaml
+++ b/charts/plex-media-server/Chart.yaml
@@ -22,7 +22,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.6.0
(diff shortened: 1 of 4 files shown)


cilindrox (approved):
LGTM 👍 - thanks @matlaj 

Going to cherrypick this into the upcoming v1.7 of the chart.

PR #181, opened 2026-08-27
fix: compare PLEX_UPDATE_CHANNEL numerically

lbellows:
## Summary

`[ "${PLEX_UPDATE_CHANNEL}" > 0 ]` is a redirect, not a comparison. When the variable is set it created a file named `0` in the working directory and treated any non-empty value as true.

Use a numeric `-gt` test (non-numeric values fail closed) and quote `PLEX_CLAIM_FILE` when reading the claim token.

## Test plan

- [x] `bash -n root/plex-common.sh root/etc/cont-init.d/40-plex-first-run`
- [x] Reproduced the old `[ > 0 ]` creating a file named `0`
- [x] New `-gt` accepts `16`, rejects `0`, fails closed on non-numeric, and does not create `0`
- [x] `PLEX_CLAIM_FILE` is quoted


cilindrox (approved):
LGTM - thanks @lbellows

PR #183, opened 2026-09-02
feat(chart): add startupProbe support to the pms container

alphayax:
## Why

The chart supports `livenessProbe` and `readinessProbe` (added in #117) but not `startupProbe`. That gap forces a bad trade-off, and the reason is already written down in #117's own review discussion:

> if it's a particularly long db migration, the restart will just result in it being run again from the beginning and will indefinitely until the migration is allowed enough time to complete.

That is exactly right, and it is the problem. PMS answers **503 on `/identity` in two completely different situations**:

1. **while it runs database migrations at startup** — transient, can take many minutes on a large library, and restarting makes it *worse* because the migration restarts from the beginning;
2. **when the server is wedged** — for example the Butler database-backup deadlock [reported on the forums](https://forums.plex.tv/t/database-backup-deadlock-with-active-library-scan-credits-detection/937113), where PMS keeps its process alive and answers 503 to everything indefinitely.

A liveness probe is the only thing that catches (2), because the process never exits. But with only a liveness probe available, its budget has to be stretched to cover the worst case of (1) — otherwise it turns a slow migration into a restart loop. So users end up either with no probe at all, or with a probe deliberately detuned to the point of being slow to help.

`startupProbe` is the Kubernetes primitive that separates the two: it suspends the liveness probe until the container answers once, then hands over. With it, the same deployment can tolerate a long migration *and* keep a tight liveness probe afterwards.

I hit this in production: PMS deadlocked on its nightly database backup and served 503 to every request for over 13 hours while the container stayed `Running`. Adding a liveness probe fixes the outage, but without a `startupProbe` I had to widen it to ~32 minutes to stay safe across version upgrades, which gives up most of the benefit.

## What this changes

Four files, and no behaviour change for anyone who does not set the new value:

- **`templates/statefulset.yaml`** — one `{{- with }}` block, mirroring the existing two exactly, placed **before** `livenessProbe`.
- **`values.yaml`** — `startupProbe: {}` with a commented-out working example, in the same style as the neighbouring two.
- **`README.md`** — helm-docs row, alphabetical position.
- **`Chart.yaml`** — `1.7.1` → `1.8.0`, matching how #168, #170 and #175 each bumped the minor version for an additive feature.

## Verification

- `helm template` with no values: output is **byte-identical** to the unpatched chart apart from the `helm.sh/chart` version label. No probe key is injected by default.
- `helm template` with `pms.startupProbe` and `pms.livenessProbe` both set: both render, correctly nested under the container, `readinessProbe` correctly absent.
- The rendered manifest passes `kubectl apply --dry-run=server` against a live cluster.

## Note on the example values

The commented example uses `periodSeconds: 10` / `failureThreshold: 180`, i.e. a 30-minute startup budget. That is deliberately generous: the cost of an over-long startup probe is a slow first start, while the cost of one that is too short is a migration restarted from scratch, repeatedly. Happy to change the numbers if you would rather the example be less conservative.


charts/plex-media-server/values.yaml
```diff
@@ -123,6 +123,18 @@ pms:
   # -- Enable process namespace sharing within the pod.
   shareProcessNamespace: false
 
+  # -- Add kubernetes startup probe to pms container. PMS answers 503 on `/identity`
+  # while it runs database migrations at startup, which can take much longer than a
+  # normal boot on a large library. A startup probe holds the liveness probe off until
+  # PMS answers once, so the liveness probe can stay tight without risking a restart
+  # loop that would relaunch the migration from the beginning.
```
cilindrox: nit: reword

```suggestion
  # -- Add a startup probe. This can be used to delay the liveness probe
  # until PMS has finished DB migrations and prevent restarts.
```

cilindrox (approved):
LGTM - thanks @alphayax - just some comments on the wording.

cilindrox:
@alphayax linter's failing - seems to be missing a `make docs`

alphayax:
@cilindrox thanks. Regenerated in 89931cd. The run is in action_required if you could approve it.

cilindrox (approved):
LGTM 👍 - thanks @alphayax

